PR #2, opened 2017-01-19
Merge branches/forks 

### LINKED ISSUE #4: Merge forks/branches
People have submitted PRs to three forks/branches recently:

1.  dask/master
2.  moody-marlin/master
3.  moody-marlin/no_api

It would be good to merge these all to some fork/branch quickly so that we don't need to go through expensive merging later on.  I don't particulaly care which fork/branch, but I would prefer to see things drive towards a common point.  It becomes expensive to merge if things diverge for too long.

cc @moody-marlin @hussainsultan

hussainsultan:
This PR merges all the work done in forks and branches that @moody-marlin would like to retain. 
- bfgs
- gradient descent
- newtons method
- proximal gradient descent
- admm for l1 logistic regression problem

I went ahead and rebased from `moody-marlin/dask-glm@no-api` branch. I also cherry picked some commits around flake8 from `moody-marlin/dask-glm@master`.

Most of the code here (except some ADMM things) is not tested and tests will be added in future PRs. I tried my best to retain as much of history as possible while squashing some commits together where it made sense.

resolves #4


hussainsultan:
Travis is failing - i need to investigate why tests are failing while they pass on my local machine.

dask_glm/logistic.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
+from scipy.optimize import fmin_l_bfgs_b
+import dask.array as da
+
+
+def logistic_regression(X, y, alpha, rho, over_relaxation):
+    N = 5
```
mrocklin: What is this `N` parameter?

dask_glm/logistic.py
```diff
@@ -0,0 +1,94 @@
...
+    a = np.zeros([n, N])
+    beta_old = da.from_array(np.zeros([n, N]), chunks=(2, 1))
+    beta = np.zeros((2, 5))
+    u = da.from_array(a, chunks=(2, 1))
+    MAX_ITER = 100
+    ABSTOL = 1e-4
+    RELTOL = 1e-2
+
+    for k in range(MAX_ITER):
+        beta_x = y.map_blocks(local_update, X, beta_old.T, z[:, 0], u.T, rho, chunks=(5, 2)).compute()
```
mrocklin: If we are only using dask.array for `map_blocks` then it might also make sense to use dask.delayed.  Keeping all arrays as numpy arrays (or lists of numpy arrays for `X` and `y`) might make things more explicit.

dask_glm/logistic.py
```diff
@@ -0,0 +1,94 @@
...
+            print("Converged!", k)
+            break
+    return z.mean(1)
+
+
+def sigmoid(x):
+    return 1 / (1 + np.exp(-x))
+
+
+def logistic_loss(w, X, y):
```
mrocklin: We might consider decorating functions like this with `numba.jit` if available

mrocklin: (Although for this particular function doesn't seem to help much (presumably most of the cost is in the `dot` call.)

```python
In [1]: def logistic_loss(w, X, y):
   ...:     y = y.ravel()
   ...:     z = X.dot(w)
   ...:     yz = y * z
   ...:     idx = yz > 0
   ...:     out = np.zeros_like(yz)
   ...:     out[idx] = np.log(1 + np.exp(-yz[idx]))
   ...:     out[~idx] = (-yz[~idx] + np.log(1 + np.exp(yz[~idx])))
   ...:     out = out.sum()
   ...:     return out
   ...: 

In [2]: import numpy as np

In [3]: X = np.random.random((1000000, 20)); y = np.random.random(1000000); w = np.random.random(
   ...: 20)

In [4]: %time logistic_loss(w, X, y)
CPU times: user 172 ms, sys: 220 ms, total: 392 ms
Wall time: 105 ms
Out[4]: 210159.75275204808

In [5]: import numba

In [6]: fast_logistic_loss = numba.jit(logistic_loss)

In [7]: %time fast_logistic_loss(w, X, y)  # first time incurs compilation cost
CPU times: user 324 ms, sys: 216 ms, total: 540 ms
Wall time: 302 ms
Out[7]: 210159.75275204808

In [8]: %time fast_logistic_loss(w, X, y)  # subsequent times are faster
CPU times: user 132 ms, sys: 212 ms, total: 344 ms
Wall time: 90.3 ms
Out[8]: 210159.75275204808

In [9]: %time fast_logistic_loss(w, X, y)  # subsequent times are faster
CPU times: user 132 ms, sys: 216 ms, total: 348 ms
Wall time: 90.9 ms
Out[9]: 210159.75275204808
```

dask_glm/logistic.py
mrocklin: Separately, it might be nice to allow this function to be used with numpy inputs rather than require that the inputs be dask.arrays.

dask_glm/logistic.py
hussainsultan: N is the number of equal sized chunks. This needs to be calculated with the dask array.

dask_glm/logistic.py
hussainsultan: If using delayed, would that mean abandoning dask.array as an input? I think it makes sense but just want to make sure i am reading it correctly.

dask_glm/logistic.py
hussainsultan: I like the idea of using numba here. I would hold back on it until we understand computational bottleneck before optimizing for speed.

hussainsultan:
@mrocklin  thanks for the comments. This is really helpful.

dask_glm/logistic.py
mrocklin: You would do something like the following:

```python
if isinstance(X, da.Array):
    X = X.to_delayed()
```

mrocklin: This really doesn't matter.  The two methods are equivalent.  It's just a question as to what is easier for people to reason about.  If you prefer dask.array then that's probably best.

mrocklin:
Heh, that's quite a push :)

Note that @moody-marlin 's revert commits can now be removed.  We've merged the persist PR into Dask.

Also, in moody-marlin/dask-glm@master there were some flake8 commits by an external contributor.  They may have some value.

hussainsultan:
@mrocklin yes! i am just making sure all test pass before i consolidate some of the commits.

hussainsultan:
@moody-marlin @mrocklin  could you please provide a review? All tests are passing on python 2 and 3. Thank you!

dask_glm/logistic.py
```diff
@@ -0,0 +1,380 @@
...
+            rhok = 1 / yk.dot(sk)
+            adj = np.eye(p) - rhok * dot(sk, yk.T)
+            Hk = dot(adj, dot(Hk, adj.T)) + rhok * dot(sk, sk.T)
+
+        step = dot(Hk, gradient)
+        steplen = dot(step, gradient)
+        Xstep = dot(X, step)
+
+        Xbeta, gradient, func, steplen, step, Xstep = da.compute(
+            Xbeta, gradient, func, steplen, step, Xstep)
```
mrocklin: This line should be removed.

mrocklin: We want to avoid ever calling `compute` on a large result (this causes cluster-to-client communication).  Instead we're starting to use persist

mrocklin: This is the only function that we have optimized in this way so far

mrocklin:
If we can merge this soon that would allow me to try optimizing a few other algorithms (I have some time today).

dask_glm/logistic.py
hussainsultan: good catch. i will remove it in a separate commit.

hussainsultan:
@mrocklin feel free to merge it once the tests pass. Thank you!

mrocklin:
I think that @moody-marlin should probably make the final call on this.

dask_glm/logistic.py
moody-marlin: Yea, and we should consolidate our loss / gradient / hessian calculations.  This immediately brings up API considerations, so I'll leave that for the near-future.

dask_glm/logistic.py
```diff
@@ -0,0 +1,380 @@
...
+    a = np.zeros([n, N])
+    beta_old = da.from_array(np.zeros([n, N]), chunks=(2, 1))
+    beta = np.zeros((2, 5))
+    u = da.from_array(a, chunks=(2, 1))
+    MAX_ITER = 100
+    ABSTOL = 1e-4
+    RELTOL = 1e-2
+
+    for k in range(MAX_ITER):
+        beta_x = y.map_blocks(local_update, X, beta_old.T, z[:, 0], u.T, rho,
```
hussainsultan: Thanks. Could we open issue for this so i can handle it in a separate PR? Thank you

moody-marlin: See #8

Issue #8, opened 2017-01-26
ADMM shape changes


moody-marlin:
A few shape issues that arose in #2 : 

- Change the ADMM implementation to accept `y` arrays of shape `(X.shape[0], )`, possibly by simply adding the line `y = y[:, None]` before restructuring too much of the code.
- Change the ADMM implementation to accept `X, y` arrays that are not split into 5 chunks, but an arbitrary number of chunks / arbitrary chunk distribution.

@hussainsultan

moody-marlin:
Resolved with PR #22

PR #10, opened 2017-01-26
Performance improvements 

mrocklin:


dask_glm/logistic.py
```diff
@@ -180,8 +188,8 @@ def newton(X, y, max_iter=50, tol=1e-8):
     '''Newtons Method for Logistic Regression.'''
 
     n, p = X.shape
-    beta = np.zeros(p)
-    Xbeta = X.dot(beta)
+    beta = np.zeros(p)  # always init to zeros?
```
moody-marlin: For now we can always initialize to zeros, but once we have a minimally viable API we should allow for alternate initializations (random, user-defined, etc.).

hussainsultan:
```python
from dask_glm.logistic import bfgs
X = np.random.multivariate_normal(np.zeros(beta_len), np.eye(beta_len), N)
y = sigmoid(X.dot(np.array([[1.5, -3]]).T)) + .001 * np.random.normal(
        size=(N, 1))
X,y = (da.from_array(X, chunks=(N / 5, 2)), da.from_array(y, chunks=(
        N / 5, 1))

bfgs(X,y)
```

```python
ValueError: Provided chunks are not consistent with shape
```

mrocklin:
It looks like examples from @moody-marlin have `y` as a row vector while your example uses `y` as a column vector.  I've squeezed y in the function now so your example should work.  This was probably pre-existing and should be part of a test suite in the future.

moody-marlin:
@hussainsultan see #8, I believe that is relevant to what you're seeing.

mrocklin:
I've resolved the test failure (just a flake8 issue) and merged the fix directly to master

PR #15, opened 2017-01-27
A test and clean-up

moody-marlin:
Added a high-level test for unregularized algorithms and cleaned up some files.  Current implementation of BFGS can't pass the test so that should be looked into.


dask_glm/tests/test_unregularized.py
```diff
@@ -0,0 +1,45 @@
...
+    X = X / col_sums[None, :]
+    X[:, p] = 1
+    X = da.from_array(X, chunks=(N / 5, p + 1))
+    y = make_y(X, beta=np.random.random(p + 1))
+
+    return X, y
+
+
+@pytest.mark.parametrize('opt',
+                         [newton, gradient_descent])
```
mrocklin: I think that you can add an entry here: `pytest.mark.xfail(bfgs, reason="some message")`

moody-marlin:
Ah thanks!  It appears that BFGS is terminating early with `stepSize == 0`; I believe it has something to do with the backtracking line search but I can't be sure yet. :/

dask_glm/tests/test_unregularized.py
```diff
@@ -0,0 +1,47 @@
...
+    X = X / col_sums[None, :]
+    X[:, p] = 1
+    X = da.from_array(X, chunks=(N / 5, p + 1))
+    y = make_y(X, beta=np.random.random(p + 1))
+
+    return X, y
+
+
+@pytest.mark.xfail(bfgs, reason='''
+        Early algorithm termination for unknown reason''')
```
mrocklin: As is this xfails all of the tests, not just the bfgs test. 

Instead, you want to wrap xfail around the bfgs element in the list in the line below.

moody-marlin: Ah, oops, embarrassing - too quick of an edit; thanks!

hussainsultan:
Sweet! Thanks @moody-marlin

dask_glm/tests/test_unregularized.py
```diff
@@ -0,0 +1,47 @@
...
+                          (250, 12, 90210),
+                          (95, 6, 70605)])
+def test_methods(N, p, seed, opt):
+    X, y = make_data(N, p, seed=seed)
+    coefs = opt(X, y)
+    p = sigmoid(X.dot(coefs).compute())
+
+    y_sum = y.compute().sum()
+    p_sum = p.sum()
+    print('y sum: {}'.format(y_sum))
```
hussainsultan: print noise - probably could be removed from the tests.

mrocklin:
How do we want to handle squashing vs merging?  What sort of git history do you all prefer?

moody-marlin:
@mrocklin as you can probably tell, I'm not an expert on Git best practices, but I do like the idea of squashing merges into one commit to keep things clean.

hussainsultan:
+1 on squashing commits.  In theory, you want to have a commit that passes all the tests and builds an incremental feature. 
Its also a good practice to add issue number if available in the commit as well:
`[# issue-number] Commit message`

I dont always follow this...but thought i mentioned.

hussainsultan:
merging. Thank you @moody-marlin

PR #19, opened 2017-01-27
Prox stepsize

moody-marlin:
Refactored proximal gradient stepsize computation and added test for l1-regularization.

### DIFF
--- a/dask_glm/logistic.py
+++ b/dask_glm/logistic.py
@@ -220,6 +220,32 @@ def newton(X, y, max_iter=50, tol=1e-8):
     return beta
 
 
+def compute_prox_step(beta, gradient, func, X, y, prox, stepSize=1.0):
+    '''Computes stepsize for proximal gradient method using
+    backtracking line search.'''
+
+    backtrackMult = 0.1
+
+    # Compute the step size
+    obeta = beta
+    lf = func
+    for ii in range(100):
+        beta = prox(obeta - stepSize * gradient, stepSize)
+        step = obeta - beta
+        Xbeta = X.dot(beta).compute()  # ugh
+
+        # This prevents overflow
+        if np.all(Xbeta < 700):
+            eXbeta = np.exp(Xbeta)
+            func = np.sum(np.log1p(eXbeta)) - np.dot(y, Xbeta)
+            df = lf - func
+            if df > 0:
+                break
+        stepSize *= backtrackMult
+
+    return stepSize, beta, Xbeta, func
+
+
 def proximal_grad(X, y, reg='l2', lamduh=0.1, max_steps=100, tol=1e-8):
     def l2(x, t):
         return 1 / (1 + lamduh * t) * x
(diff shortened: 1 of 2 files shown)


moody-marlin:
cc @mrocklin 

I think there will be further performance gains with well-placed computes / persists that we haven't done for `proximal_grad` yet, but this pulls out the stepsize computation and adds a test (@hussainsultan ;) )

dask_glm/tests/test_regularized.py
```diff
@@ -0,0 +1,38 @@
...
+
+    return X, y
+
+
+@pytest.mark.parametrize('opt', [proximal_grad])
+@pytest.mark.parametrize('N, p, seed',
+                         [(100, 2, 20009),
+                          (250, 12, 90210),
+                          (95, 6, 70605)])
+def test_l1(N, p, seed, opt):
```
hussainsultan: nit picky thing - could we change the name of the method from `test_l1` to something more meaningful? Perhaps `test_prox_grad_with_large_alpha`?

mrocklin:
The function fails to run currently (missing reference to `nextBacktrackMult`).

When I remove this reference it does run, but "converges" to values that around 10% apart from each other on repeated runs.

moody-marlin:
Ah this is actually a great catch; for large values of `alpha` we expect `proximal_grad` with `reg = 'l1'` to return a coefficient vector of zeros, but that is *also* the starting value for the algorithm, so I didn't catch the missing reference which appears *after* the convergence check.  Because of this, for this test there should be random initializations.

dask_glm/logistic.py
```diff
@@ -220,6 +220,32 @@ def newton(X, y, max_iter=50, tol=1e-8):
     return beta
 
 
+def compute_prox_step(beta, gradient, func, X, y, prox, stepSize=1.0):
```
mrocklin: This function takes in `X` which seems like it would affect scalability.

dask_glm/logistic.py
```diff
@@ -220,6 +220,32 @@ def newton(X, y, max_iter=50, tol=1e-8):
...
+
+    backtrackMult = 0.1
+
+    # Compute the step size
+    obeta = beta
+    lf = func
+    for ii in range(100):
+        beta = prox(obeta - stepSize * gradient, stepSize)
+        step = obeta - beta
+        Xbeta = X.dot(beta).compute()  # ugh
```
mrocklin: In bfgs we compute an XStep and then push that forward.  Is that possible here as well?

moody-marlin: Unfortunately not; the beta update in proximal gradient descent is non-linear, so the new `Xbeta` can't be computed purely from `obeta` and `Xstep` (at least not that I am aware of).  Maybe @mcg1969 has some input about the best way to do this, avoiding repeated matrix-vector operations within the line search?

mcg1969: That's right, but again, the number of backtracks should be small after the first iteration or two.

mrocklin: OK, in that case I'll take back my request that we pull this function out.  It seems like all of the line search should happen with Dask and not in a single thread, correct?

PR #20, opened 2017-01-28
Add basic convergence tests

mrocklin:
This adds a parametrized test to ensure basic functionality of all solvers

I am somewhat concerned about the lack of accuracy in the outputs.  I had to
crank the tolerance very low on final testing.  But perhaps this is to be
expected given the small datasets?

This may also serve as a model for testing in the future?

### DIFF
--- a/dask_glm/tests/test_logistic.py
+++ b/dask_glm/tests/test_logistic.py
@@ -11,9 +11,9 @@ from dask_glm.utils import make_y
 @pytest.mark.parametrize('func,kwargs', [
     (newton, {'tol': 1e-3}),
     (bfgs, {'tol': 1e-8}),
-    (gradient_descent, {'tol': 1e-5}),
-    (proximal_grad, {'tol': 1e-5, 'reg': 'l1'}),
-    (proximal_grad, {'tol': 1e-5, 'reg': 'l2'}),
+    (gradient_descent, {'tol': 1e-6}),
+    (proximal_grad, {'tol': 1e-6, 'reg': 'l1'}),
+    (proximal_grad, {'tol': 1e-6, 'reg': 'l2'}),
 ])
 @pytest.mark.parametrize('N', [10000, 100000])
 @pytest.mark.parametrize('nchunks', [1, 10])
--- a/dask_glm/tests/test_logistic.py
+++ b/dask_glm/tests/test_logistic.py
@@ -12,8 +12,8 @@ from dask_glm.utils import make_y
     (newton, {'tol': 1e-5}),
     (bfgs, {'tol': 1e-8}),
     (gradient_descent, {'tol': 1e-7}),
-    (proximal_grad, {'tol': 1e-6, 'reg': 'l1'}),
-    (proximal_grad, {'tol': 1e-7, 'reg': 'l2'}),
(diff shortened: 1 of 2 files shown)


mrocklin:
cc @hussainsultan @moody-marlin

moody-marlin:
The proximal methods (in this case just `proximal_grad`) *shouldn't* converge to the "true" `beta` values; the regularization "shrinks" the coefficients.  

I'll check the tolerance settings you have in this test, but default settings are typically around ~1e-8 so it doesn't appear that these are overly strict.

mrocklin:
I was more concerned about the rtol value in np.assert_equals at the end.
We're getting relatively imprecise answers.  Regularization could be the
cause though.

mrocklin:
How should we handle randomized tests?  Use seeds?  Use tolerances and situations such that failures are very very unlikely?

cc @hussainsultan it looks like some of your tests fail randomly as well.

moody-marlin:
We could generate a handful of random datasets and use those exclusively?  My initial plan was to use tolerances and failures that are unlikely, but I realize that is somewhat brittle.

mrocklin:
To me random datasets that we use exclusively is equivalent to "use random seeds".  Yes?

mrocklin:
I'm also fine with having very low probability failures.  Presumably this is easier if we use larger datasets which should become less onerous as things become faster.

mrocklin:
I've tweaked tolerance settings until they pass repeatedly on my machine.  Merging soon if no objections.

hussainsultan:
Looks good to me. Thank you!

moody-marlin:
@mrocklin this is not the right test for `proximal_grad`, it should be removed from the list; if you want, you can add in the regularized test I have in my PR #19 for `proximal_grad`.

mrocklin:
What is wrong with the proximal grad test here?  It there a way to include proximal grad in a large parametrized test like this?  What would we have to change to include it.

I find these things useful because as we add things (like say distributed computing) we can add a single parameter to this test and reliably hit all implementations.

moody-marlin:
Regularized problems are not necessarily asymptotically consistent; this means that the "correct" solution to a regularized problem is not going to be the "true" underlying `beta` that generated the data.  (Which is what this test is testing for).

Even in the limit of larger datasets, the optimal regularized coefficients will be smaller in size than the "true" betas, but how much smaller I can't tell you _a priori_ (which is why regularized optimization can be hard / why some old-school statisticians don't like them).

For testing the regularized methods we will either have to stick to high-level tests, or generate a few *fixed* datasets that we definitely know the solutions for, and test against those.

mrocklin:
Can we add a flag for regularized tests so that we accept a basic extra error tolerance?  My goal here isn't so much to test full accuracy, but just to make sure that we run *everything* and get some sort of sensible answer back.

moody-marlin:
Yea, that's fair; I think if you test `proximal_grad` with `lamduh = 0.001` or something small, then the test with `1e-1` tolerance should be OK going forward.

moody-marlin:
FWIW I think the high-level tests are still useful for added confidence.

mrocklin:
I've added in the lamduh term.  Good thinking.  Other comments?

moody-marlin:
I think this should be good.

PR #21, opened 2017-01-30
Use persist in proximal grad

mrocklin:
This appears to provide a substantial speedup in single-machine processing as well.

cc @moody-marlin


mrocklin:
Merging soon if no further comments

moody-marlin:
Looks good to me!

Issue #16, opened 2017-01-27
Add default arguments to ADMM code


mrocklin:
I started trying to play with @hussainsultan 's ADMM code.  However, it requires many parameters for which I do not know good values.  Can we add sane defaults there?  Do we know sane defaults?

Short term it would be useful to me to always be able to do the following for all solver routines:

    result = func(X, y)

hussainsultan:
Thanks for the call out. Thats sometime that i intended to clean in my initial PR but then got sidetracked by consolidation. I will take this on.

moody-marlin:
This should be resolved in PR #22

hussainsultan:
resolved by #22. closing

PR #22, opened 2017-01-31
ADMM

moody-marlin:
Using @hussainsultan 's code as a starting point I rewrote the ADMM implementation using `dask.delayed` instead of `map_blocks`, avoiding some of the shape issues.  Added in a few tests as well.


moody-marlin:
Also, resolves #16 #8

dask_glm/logistic.py
```diff
@@ -288,97 +288,92 @@ def identity(x, t):
...
+    yD = y.to_delayed().flatten().tolist()
+
+    z = np.array([np.zeros(p) for i in range(chunks)])
+    u = np.array([np.zeros(p) for i in range(chunks)])
+    betas = np.array([np.zeros(p) for i in range(chunks)])
+
+    for k in range(max_iter):
+
+        # x-update step
+        new_betas = [delayed(local_update)(xx, yy, bb, zz, uu, rho) for
```
hussainsultan: this is so nice! Thanks you!

dask_glm/logistic.py
```diff
@@ -288,97 +288,92 @@ def identity(x, t):
...
+                     xx, yy, bb, zz, uu in zip(XD, yD, betas, z, u)]
+        new_betas = np.array(da.compute(*new_betas))
+
+        #  z-update step
```
hussainsultan: it appears that `over_relaxation` parameter got dropped from the algorithm. Could this be added back?

dask_glm/tests/test_admm.py
```diff
@@ -0,0 +1,40 @@
...
+
+from dask import persist
+import dask.array as da
+import numpy as np
+
+from dask_glm.logistic import admm, local_update
+from dask_glm.utils import make_y
+
+
+@pytest.mark.parametrize('N', [1000, 10000])
```
hussainsultan: For just `local_update` test, this maybe an overkill. Thoughts? ` test_admm_with_large_lamduh` should capture the high level tests correct?

dask_glm/logistic.py
moody-marlin: Yup, added back.

dask_glm/tests/test_logistic.py
```diff
@@ -4,8 +4,46 @@
...
+                          (250, 12, 90210),
+                          (95, 6, 70605)])
+def test_methods(N, p, seed, opt):
+    X, y = make_data(N, p, seed=seed)
+    coefs = opt(X, y)
+    p = sigmoid(X.dot(coefs).compute())
+
+    y_sum = y.compute().sum()
+    p_sum = p.sum()
+    assert np.isclose(y.compute().sum(), p.sum(), atol=1e-1)
```
mrocklin: Does this overlap with the parametrized test just below?

dask_glm/logistic.py
```diff
@@ -288,97 +288,94 @@ def identity(x, t):
...
+         max_iter=100, abstol=1e-4, reltol=1e-2):
+
+    chunks = X.npartitions
+    (n, p) = X.shape
+    XD = X.to_delayed().flatten().tolist()
+    yD = y.to_delayed().flatten().tolist()
+
+    z = np.array([np.zeros(p) for i in range(chunks)])
+    u = np.array([np.zeros(p) for i in range(chunks)])
+    betas = np.array([np.zeros(p) for i in range(chunks)])
```
mrocklin: `np.zeros((chunks, p))`

Also, slight preference to use the term `nchunks` to avoid colliding with dask.array's use of chunks, which has a different meaning.

dask_glm/logistic.py
```diff
@@ -288,97 +288,94 @@ def identity(x, t):
...
+        new_betas = np.array(da.compute(*new_betas))
+
+        beta_hat = over_relax * new_betas + (1 - over_relax) * z
+
+        #  z-update step
         zold = z.copy()
-        ztilde = np.mean(beta_hat + a, 1)
+        ztilde = np.mean(beta_hat + np.array(u), axis=0)
+        z = shrinkage(ztilde, lamduh / (rho * chunks))
+        z = np.array([z for i in range(chunks)])
```
mrocklin: Is `z` actually 2d or are you just broadcasting out for convenience?

dask_glm/tests/test_admm.py
moody-marlin: You're probably right; I can remove it in a bit.

dask_glm/tests/test_logistic.py
moody-marlin: Yes, at the moment this test should pass if the parameterized test passes, but I think it will be useful to have when we move to testing poorly conditioned input data; in that case our parameter estimates might be hard to get accurately (because there's ~a whole linear subspace of them) but the predictions should be the same no matter which one we land on (which is what this is basically testing).

dask_glm/logistic.py
moody-marlin: Good call out; updated.

dask_glm/logistic.py
moody-marlin: Just broadcasting out for convenience; now that you point it out I see there isn't a need to replicate `z`.  Will update.

moody-marlin: Updated to rely on `numpy` broadcasting for `z`.

mrocklin:
I'm going to take a crack at this.  Merging for now.

Issue #23, opened 2017-02-01
Approximate exponents


mrocklin:
Fun fact: the current ADMM implementation can spend almost half of it's time computing `np.exp`

### Script

```python
from dask import persist
import dask.array as da
import numpy as np
from dask_glm.logistic import admm
from dask_glm.utils import make_y

N = 1e7
M = 2
chunks = 1e6
seed = 20009

X = da.random.random((N, M), chunks=(chunks, M))
y = make_y(X, beta=np.array(list(range(M))), chunks=chunks)

X, y = persist(X, y)

%%prun
import dask
with dask.set_options(get=dask.get):
    beta = admm(X, y)
```

### Profile results

I use a wrapped version of `np.exp` just so that it shows up in profile results.

```
         565331 function calls (532847 primitive calls) in 93.202 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2254   41.543    0.018   41.543    0.018 utils.py:25(exp)
     1127   30.614    0.027   52.086    0.046 logistic.py:349(logistic_loss)
     1127   13.122    0.012   40.358    0.036 logistic.py:364(logistic_gradient)
     1127    6.267    0.006   27.191    0.024 logistic.py:343(sigmoid)
     1130    0.793    0.001    0.793    0.001 {method 'reduce' of 'numpy.ufunc' objects}
        8    0.078    0.010    0.079    0.010 executionengine.py:100(finalize_object)
     1127    0.052    0.000   40.409    0.036 logistic.py:373(proximal_logistic_gradient)
     1127    0.049    0.000   52.135    0.046 logistic.py:358(proximal_logistic_loss)
        8    0.043    0.005    0.043    0.005 passmanagers.py:94(run)
       30    0.034    0.001   93.040    3.101 lbfgsb.py:205(_minimize_lbfgsb)
     2254    0.031    0.000   41.582    0.018 dispatcher.py:152(__call__)
     1127    0.029    0.000   92.998    0.083 lbfgsb.py:277(func_and_grad)
26957/12971    0.027    0.000    0.083    0.000 {method 'format' of 'str' objects}
     1127    0.021    0.000   52.417    0.047 optimize.py:290(function_wrapper)
    80776    0.019    0.000    0.030    0.000 {built-in method isinstance}
     1127    0.013    0.000    0.814    0.001 fromnumeric.py:1743(sum)
```

### Comments

Here are some comments from @stuartarchibald

[...]

PR #24, opened 2017-02-01
Test correctness of admm

mrocklin:
I went through ADMM for performance reasons but this algorithm seemed fine as is.  Apparently this has very little data transfer, so no persist magic seemed to be useful.  

On profiling I did find that `exp` issues were pretty serious here (see #23). 

I also extended the existing super-test to include admm and fixed the output

cc @moody-marlin

### DIFF
--- a/dask_glm/logistic.py
+++ b/dask_glm/logistic.py
@@ -336,7 +336,7 @@ def admm(X, y, lamduh=0.1, rho=1, over_relax=1,
             print("Converged!", k)
             break
 
-    return z.mean(0)
+    return z
 
 
 # TODO: Dask+Numba JIT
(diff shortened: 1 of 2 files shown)


dask_glm/logistic.py
```diff
@@ -336,7 +336,7 @@ def admm(X, y, lamduh=0.1, rho=1, over_relax=1,
             print("Converged!", k)
             break
 
-    return z.mean(0)
+    return z
```
hussainsultan: is z a 2D array here?

mrocklin: It used to be, now it's 1d.

hussainsultan: Thats so nice. Thanks!

PR #25, opened 2017-02-02
Parallelize stepsize computation

mrocklin:
Previously computation time in gradient descent was blocked by the sequential stepsize calculation.  It turns out that parallelizing this isn't that bad, at least not when latency to the cluster is low.

Sequential version plot: https://cdn.rawgit.com/mrocklin/163be5f81a446931e49d4382e6ae720f/raw/57156dd63720f57b471e036f7e94ee6ff8a8d6d0/dask-glm-stepsize-sequential.html

Parallel version plot: https://cdn.rawgit.com/mrocklin/163be5f81a446931e49d4382e6ae720f/raw/63ac2ff2e4bc917be4d8e8fa6d1cf4736e98217a/dask-glm-stepsize-parallel.html

Speedups on a simple problem are around 30%

(note, both commits here are valuable for git history, recommend rebase-and-merge)

cc @moody-marlin

### DIFF
--- a/dask_glm/logistic.py
+++ b/dask_glm/logistic.py
@@ -151,18 +151,24 @@ def gradient_descent(X, y, max_steps=100, tol=1e-14):
         gradient = X.T.dot(eXbeta / e1 - y)
         Xgradient = X.dot(gradient)
 
-        Xbeta, eXbeta, func, gradient, Xgradient = da.compute(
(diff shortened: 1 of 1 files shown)


mrocklin:
Updated plots in the main comment to point to the right profiles (they're worth checking out)

Issue #11, opened 2017-01-26
API discussion


cicdw:
One of the biggest hanging issues is to organize / discuss what the **API** will look like for `dask_glm` and what sorts of things we expose.  This decision impacts algorithm structure going forward (as we incorporate other GLM families and regularizers) so I will need some help focusing the following discussion on specific action items.  My initial thoughts were to:

- *Aim for generality so users can create their own GLM families / regularizers*: My initial thought along these lines were to have an `Optimizer(**initialization_args, **numerical_algo_args, **regularization_args)` class that can be inherited by any class that implements `func(), grad(), hessian()` methods.  This would allow us to focus on algorithm development separately from model specification and add additional model types (Normal, Poisson, etc.) easily.  The user then calls `Logistic(X,y, **kwargs)` to initialize a model, and then `.fit(**algo_args)` calls on the methods from `Optimizer`.  This can get hairy, and could possibly prevent further optimizations that exploit specific structure inherent in the models.

- *Hard code the models users can choose from*: Per my conversation with @jcrist, we might not want to aim for extreme flexibility in the model API, as most people probably don't need / want to define their own GLM families or their own regularizers.  Related, we might want to implement the algorithms specifically for each family / regularizer.

- *Regularized vs. unregularized models*: How do we manage the 'switch' from regularized to unregularized models?  Should it be something as simple as `model = LogisticModel(X, y, reg='l2')` and the user chooses from a pre-defined list of regularizers, or should we allow for `model = LogisticModel(X, y, reg=reg_func)` where `reg_func()` is any regularization function on the parameters, allowing for customizations such as grouped lasso, non-negativity, etc?  Again I was hoping to keep things as general as possible so that the user can mix and match, but on the algorithm side this can get difficult especially if we want speed and accuracy.

-*Inherit from scikit-learn*: Per my conversation with @hussainsultan, we might want to try and stick as closely as possible to `scikit-learn`'s API, as this is common enough to be comfortable for most people.  However, we don't want to adhere to this too strongly (especially in output), because we started with the goal of intelligently toggling between inferential and predictive applications (thus requiring things like inferential stats that `scikit-learn` currently doesn't include).

Ultimately, I think my thought process boils down to: *should we expose an API to access the underlying algorithms or should we bake them into the various specific GLM families we want to consider?*

cc: @mrocklin @mcg1969

mrocklin:
[...]

PR #28, opened 2017-02-20
Algorithm independence and Accuracy Concerns

moody-marlin:
Work in progress! Most of the algorithms have been made independent of logistic regression, so that we can consider easily adding other GLM families (normal model has been added here as an example).

I also wrote up a notebook in `notebooks/` that points out a few accuracy issues I'm seeing so that we can discuss whether or not these are of any concern to us.  One of the biggest reasons this worries me is that it makes test writing hard; if we were finding really good optima then we can easily test for that, but it seems that not all of the algorithms are finding high quality points which has been a bottleneck for writing robust tests.

Related to this, it'd be nice to set a rule for our tolerance settings / test requirements -- something like: "For any algorithm, no matter the convergence criterion, at tolerance 1e-8 we should have agreement with the true solution to 4 decimal places."  or something.


moody-marlin:
cc: @mcg1969 would be interested in your thoughts on the notebook I included in this PR.

mrocklin:
At first glance this looks pretty slick to me.  I now understand your desire to have a class with a fixed set of staticmethods.

hussainsultan:
Looks pretty sweet!

dask_glm/normal.py
```diff
@@ -0,0 +1,36 @@
...
+    Xbeta = X.dot(beta)
+    return gradient(Xbeta, X, y)
+
+
+def gradient(Xbeta, X, y):
+    return 2 *  dot(X.T, Xbeta) - 2 * dot(X.T, y)
+
+
+def hessian(Xbeta, X):
+    return dot(X.T, X)
```
mrocklin: In conversation I originally pushed for functions like newton, gradient_descent, etc.. to just take in flat functions.  At the time I didn't realize how similarly structured the different approaches here would be.  I'd like to relax my request to just that those functions should take in something simple, that is hopefully separated from other concerns like "should we use the statsmodels or sklearn APIs".  It may be that constructing a simple class around these functions is best.  

Three options come to mind:

1.  A class with these functions as static methods.  We would never instantiate the class, but just use it as a namespace
2.  A class with these functions as methods.  This would be useful if future options other than Linear and Logistic regression are likely to have parameters that we would want to tune
3.  Leave it as is, but use the modules themselves as inputs
4.  Leave it as is, and continue passing in functions directly

mrocklin:
Are there ways to reasonably test all solvers against all problem types like logistic/linear/etc.?  Perhaps something that is parameterized so that we don't have to keep writing new tests every time we extend?

mrocklin:
Even a very small/dumb test on very simple data that ran on everything would probably do a great deal of good.  This wouldn't even have to test accuracy in a reasonable way.  Its goal would be to just run all of the lines of code to catch any egregious mistakes.

dask_glm/normal.py
moody-marlin: After thinking over @mcg1969's [comment on the API](https://github.com/dask/dask-glm/issues/11#issuecomment-281191366), I'm on board with having a class with static methods that hold these functions (loglike, gradient, hessian, etc.), one class for each unregularized GLM family.

Separately, we can have a "model" class (also probably one for each family) that has the fit / summary / initialization routines.

mrocklin: I'm curious, how orthogonal are all of these choices?  Is it easy to mix-and-match GLM family, solver, etc.?

mrocklin: @jcrist , this also seems like a relevant occasion for your input/time if you're interested.

mcg1969:
First-order methods are generally quite reliable at finding an accurate solution if you _completely_ ignore iteration counts. If you simply want to make sure that a particular function/gradient combo are compatible with each other, you can just let them run a long time ;-)

mcg1969:
For second-order methods, there's a bit of a complication. The Newton direction is guaranteed to be a descent direction in exact arithmetic, but if the Hessian becomes more poorly conditioned, this guarantee may fail to hold in finite precision. It would be good to consider a sanity check of some sort to ensure that `<g, H^{-1} g> < 0`.

It is also _very often_ the case that someone gets the Hessian calculation wrong, and this error manifests itself in subtle ways.

dask_glm/normal.py
mcg1969: I have found it beneficial in the past to have a mathematical function be represented as a class that accepts the input value as state, in order that intermediate calculations can be re-used across function, gradient, and Hessian calculations.

mcg1969: If you prefer a stateless approach, then a function that can return 1, 2, or 3 outputs depending on the number of derivatives needed would work.

mcg1969:
Here's some code I wrote for another purpose:
```
class Logistic(object):
    def __init__(self, y, verbose=False):
        try:
            y = np.asarray(y, dtype=np.bool)
        except:
            raise ValueError('y must be convertible to a boolean array')
        if y.ndim > 1:
            raise ValueError('y must be one-dimensional')
        elif y.ndim == 0:
            y = np.reshape(y, (1,))
        self._y  = y
        self._ex = np.empty(y.shape)
        self._f  = np.empty((1,))
        if verbose:
            print('Logistic objective: length %d.' % y.size)

    @property
    def size(self):
        return self._y.size

    def f(self, x):
        return calc_f(x, self._ex, self._y)

    def g(self, out=None):
        return calc_g2(self._ex,self._y,out=out)

    def Hv(self, v, out=None):
        return calc_H2(v,self._ex,out=out)
```

dask_glm/normal.py
moody-marlin: @mcg1969 Backtracking on my "all static methods" comment, I agree that preserving state actually makes sense here; for example, with a Gaussian / normal family, the Hessian only should be computed *once* because it is independent of `beta`.  Similarly, `p = sigmoid(Xbeta)` is re-used across gradient / Hessian calculations in the logistic model.

@mrocklin Once a log-likelihood is chosen, the gradient / hessian are determined for you -- which solver you use is *fairly* independent of family, as long as the model has enough derivatives for the solver to function properly (example: adding in `l1` regularization restricts you to `proximal_grad` and `admm` for the moment).

moody-marlin: Once we have an interface for adding in ~arbitrary / custom regularizers, then the mixing and matching will get more interesting.

mrocklin: Mostly my objective is to ensure that we don't end up having to create `LogisticADMM` and `LinearGradientDescent` classes, but can instead pull a few orthogonal things together.

mrocklin: It sounds like you're on top of this though.

moody-marlin:
Refactored to a `families.py` file for holding the classes with static methods; I want to think more about how to better utilize repeated calculations by maintaining state within the classes, but this is at least a first step to the generalization we are aiming for.  The algorithms still currently take in individual functions instead of a class name / instance.  Ready to merge pending someone else's approval / review.

cc: @mrocklin @hussainsultan

mrocklin:
Github has concerns.  There is a merge conflict with upstream.

mrocklin:
In cases where the functions operate on dask.array objects (gradient descent, newton. proximal) you don't need to worry about sharing intermediates explicitly.  Dask will take care of this for you.  This won't carry over to ADMM though where you're just applying the functions directly onto numpy arrays.

dask_glm/families.py
```diff
@@ -0,0 +1,70 @@
...
+from dask_glm.utils import dot, exp, log1p, sigmoid
+
+
+try:
+    from numba import jit
+except ImportError:
+    def jit(*args, **kwargs):
+        def _(func):
+            return func
+        return _
```
mrocklin: This code now appears to be unused.

dask_glm/algorithms.py
```diff
@@ -126,7 +42,8 @@ def compute_stepsize(beta, step, Xbeta, Xstep, y, curr_val, stepSize=1.0,
     return stepSize, beta, Xbeta, func
 
 
-def gradient_descent(X, y, max_steps=100, tol=1e-14):
+def gradient_descent(X, y, max_steps=100, tol=1e-14, loglike=Logistic.loglike,
+                     gradient=Logistic.gradient):
```
mrocklin: Should these functions now take in a class instead of several staticmethods?

dask_glm/families.py
moody-marlin: Removed.

mrocklin:
Hooray merge conflicts resolved!

mrocklin:
Ah, although we've now overwritten the changes in https://github.com/dask/dask-glm/pull/25 that allowed us to parallelize the stepsize computation.

moody-marlin:
Resolving merge conflicts is my least favorite thing; no!  Did it happen with this commit: https://github.com/dask/dask-glm/pull/28/commits/2b24b19a067f1f72c3c35839e9b426ce2cf0153b?

mrocklin:
I suspect it happened far earlier.  I suspect that the following events occurred

1.  You branched off from master, moving logistic.py to algorithms.py and making some of your own changes
2.  We merged in https://github.com/dask/dask-glm/pull/25 , significantly changing logistic.py
3.  You submitted this PR, which has all of the old versions

mrocklin:
I recommend starting from the code after https://github.com/dask/dask-glm/pull/25 was merged and seeing what would have to be done to add in the abstract stuff on top of it and then replacing your current solution with that.  If this ends up taking a long time let me know and I'll see if I can accelerate things.

moody-marlin:
@mrocklin OK I think I fixed it; tests are passing.

.coverage
```diff
@@ -0,0 +1 @@
...62],"/Users/nwl814/Documents/Python/dask-glm/dask_glm/families.py":[64,1,3,68,6,7,15,17,19,20,22,25,26,27,29,32,33,34,36,38,39,41,43,44,47,48,52,58],"/Users/nwl814/Documents/Python/dask-glm/dask_glm/algorithms.py":[1,3,4,5,6,7,10,11,14,15,23,24,25,26,27,28,29,30,31,32,33,34,36,37,38,39,40,42,45,46,49,50,51,52,53,54,55,56,57,58,60,62,63,64,66,67,69,70,73,74,75,76,77,78,79,80,81,85,86,88,89,90,91,92,94,97,100,101,102,104,105,107,108,111,112,114,118,119,121,124,126,128,129,131,134,135,136,138,139,140,141,143,144,145,147,148,150,153,155,156,158,161,162,163,166,169,170,172,173,174,175,177,178,179,181,184,185,187,188,191,192,194,195,196,199,200,202,203,204,205,206,207,208,210,213,214,215,218,221,222,224,225,226,227,228,230,231,232,234,235,236,237,239,240,242,243,244,245,246,248,249,250,253,254,255,256,257,258,259,260,261,262,263,265,266,268,270,271,273,274,275,278,280,281,282,284,287,288,289,290,291,293,296,297,298,300,301,303,306,307,308,309,310,311,312,313,314,315,317,321,323,324,325,326,327,328,330,331,333,334,337,338,339,340,341,343,344,345,348,349,350,351,352,353,354,355,356,357,360,361,362,364,365,366,367,368,370],"/Users/nwl814/Documents/Python/dask-glm/dask_glm/__init__.py":[1]}}
```
mrocklin: We might want to add this file to .gitignore

dask_glm/algorithms.py
```diff
@@ -159,42 +72,41 @@ def gradient_descent(X, y, max_steps=100, tol=1e-14):
...
 
         # backtracking line search
         lf = func
-        stepSize, _, _, func = compute_stepsize_dask(beta, gradient,
+        stepSize, _, _, func = compute_stepsize_dask(beta, grad,
                                                      Xbeta, Xgradient,
                                                      y, func,
                                                      backtrackMult=backtrackMult,
                                                      armijoMult=armijoMult,
                                                      stepSize=stepSize)
```
mrocklin: We should pass down the family here

dask_glm/algorithms.py
```diff
@@ -1,155 +1,68 @@
...
         stepSize *= backtrackMult
 
     return stepSize, beta, Xbeta, func
 
-
-def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val, stepSize=1.0,
-                          armijoMult=0.1, backtrackMult=0.1):
-    beta, step, Xbeta, Xstep, y, curr_val = persist(beta, step, Xbeta, Xstep, y, curr_val)
+def compute_stepsize(beta, step, Xbeta, Xstep, y, curr_val, loglike=Logistic.loglike,
+                     stepSize=1.0, armijoMult=0.1, backtrackMult=0.1):
```
mrocklin: Thoughts on just always passing the family around everywhere rather than the staticmethods?

dask_glm/algorithms.py
```diff
@@ -159,42 +72,41 @@ def gradient_descent(X, y, max_steps=100, tol=1e-14):
     recalcRate = 10
     backtrackMult = firstBacktrackMult
     beta = np.zeros(p)
+    y_local = y.compute()
```
mrocklin: This shouldn't be necessary

dask_glm/algorithms.py
```diff
@@ -159,42 +72,41 @@ def gradient_descent(X, y, max_steps=100, tol=1e-14):
...
+            func = loglike(Xbeta, y)
 
-        e1 = eXbeta + 1.0
-        gradient = X.T.dot(eXbeta / e1 - y)
-        Xgradient = X.dot(gradient)
+        grad = gradient(Xbeta, X, y)
+        Xgradient = X.dot(grad)
+
+        Xbeta, func, grad, Xgradient = da.compute(
+            Xbeta, func, grad, Xgradient)
```
mrocklin: This shouldn't be necessary

dask_glm/algorithms.py
moody-marlin: So also for `compute_stepsize`? That's fine by me if you think that will stand up better.

mrocklin: It's a genuine question, not a request.  It might be simpler to always pass around and use the one thing.  I don't know though.

`compute_stepsize` should no longer be used anywhere I think

.coverage
moody-marlin: Done.

.gitignore
```diff
@@ -1,6 +1,7 @@
 *.pyc
 *.swpi
 *.cache
+*.coveragerc
```
mrocklin: We *do* want the coveragerc file, we don't want the coverage file.  If you look you'll find that the coveragerc file has configuration information about using coverage (a tool to test which lines are covered by tests) and that the coverage file has a bunch of data for the last particular run of the coverage tool.

moody-marlin: Haha oops, thanks for having such good eyes.

moody-marlin: I'll squash all these spurious commits after we're done cleaning it up.

dask_glm/algorithms.py
moody-marlin: I think it's a fair point; that way if we ever change what information we need / utilize a different method it would be easy to change.

moody-marlin:
`bfgs` failed, but we haven't been maintaining that recently; I can look into it further in a separate PR if we think it's worthwhile.

mrocklin:
What happened since master that would have caused this change?  Did the bfgs code change at all?

moody-marlin:
As far as I can tell with my eyes, nothing changed; not sure what's going on with that but `bfgs` appears to be returning 'nan'?

mrocklin:
Ran a diff over the two versions and came up with the following:

```diff
mrocklin@carbon:~/workspace/dask-glm$ diff foo.py bar.py 
1c1
< def bfgs(X, y, max_iter=500, tol=1e-14):
---
> def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
40c40
<                                                      y, func,
---
>                                                      y, func, family=family,
```

mrocklin:
So I'm not sure what happened.  I'll leave the decision as to what to do to you.

moody-marlin:
@mrocklin I figured it out, should be fixed now; do you suggest I update the tests to reflect all pairs now or do that in a separate PR?

moody-marlin:
Choosing to `xfail` anything involving `bfgs`; we were already doing this for some of the other tests -- I think `bfgs` will need to be looked at more closely in the near future if we want to keep it around.  

I'll do a new PR with further testing since this one has gotten fairly large.

dask_glm/tests/test_algos_families.py
```diff
@@ -0,0 +1,105 @@
...
+
+    result = func(X, y, family=family, lamduh=lam, **kwargs)
+    test_vec = np.random.normal(size=2)
+
+    f = add_l1(family.pointwise_loss, lam)
+
+    opt = f(result, X, y).compute()
+    test_val = f(test_vec, X, y).compute()
+
+    assert opt < test_val
```
mrocklin: Nice tests.

mrocklin:
This looks good to me.  +1

PR #29, opened 2017-02-22
Regularizer class

moody-marlin:
Abstract the regularizers into classes for easy adjustments later. The names of the methods might need some adjustments but this wlll also allow us to easily use all methods for any regularizer -- example, if we wrap the gradient and hessian with L2, we can run newtons method for L2-regularized problems.

(Sorry @mrocklin , I couldn't help myself doing this before the API!)

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -7,9 +7,9 @@ import dask.array as da
 from scipy.optimize import fmin_l_bfgs_b
 
 
-from dask_glm.utils import dot, exp, log1p, absolute, sign
+from dask_glm.utils import dot, exp, log1p
 from dask_glm.families import Logistic
-from dask_glm.regularizers import L1, L2
+from dask_glm.regularizers import L1
 
 
 def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -9,6 +9,7 @@ from scipy.optimize import fmin_l_bfgs_b
 
 from dask_glm.utils import dot, exp, log1p, absolute, sign
 from dask_glm.families import Logistic
+from dask_glm.regularizers import L1, L2
 
 
 def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
@@ -134,7 +135,22 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
     return beta
 
 
-def admm(X, y, lamduh=0.1, rho=1, over_relax=1,
+def add_reg_grad(func):
+    @functools.wraps(func)
+    def wrapped(beta, X, y, z, u, rho):
+        return func(beta, X, y) + rho * (beta - z + u)
+    return wrapped
+
+
+def add_reg_f(func):
+    @functools.wraps(func)
+    def wrapped(beta, X, y, z, u, rho):
+        return func(beta, X, y) + (rho / 2) * np.dot(beta - z + u,
+                                                     beta - z + u)
+    return wrapped
+
+
+def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
          max_iter=100, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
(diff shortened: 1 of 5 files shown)


dask_glm/algorithms.py
```diff
@@ -134,7 +135,22 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
...
+
+def add_reg_f(func):
+    @functools.wraps(func)
+    def wrapped(beta, X, y, z, u, rho):
+        return func(beta, X, y) + (rho / 2) * np.dot(beta - z + u,
+                                                     beta - z + u)
+    return wrapped
+
+
+def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
```
mrocklin: I still recommend allowing `reg='l1'` as an input here.

moody-marlin: I was thinking of elevating that option to the top-level API, just so I didn't have the check `if reg in ['l1', 'l2]` floating around in multiple places.

dask_glm/algorithms.py
```diff
@@ -134,12 +135,28 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
...
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
 
+    def create_local_gradient(func):
+        @functools.wraps(func)
+        def wrapped(beta, X, y, z, u, rho):
+            return func(beta, X, y) + rho * (beta - z + u)
+        return wrapped
+
+    def create_local_f(func):
```
hussainsultan: Could this be renamed perhaps? The name doesnt tell me what this function is trying to do. Feel free to ignore if this doesn't make sense at this point.

dask_glm/algorithms.py
```diff
@@ -134,12 +135,28 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
     return beta
 
 
-def admm(X, y, lamduh=0.1, rho=1, over_relax=1,
+def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
```
hussainsultan: For ADMM, would this work with other regularizers? My understanding was that this algorithm is `L1` specific.  If this is for the sake of consistent API, it makes sense to keep it.

dask_glm/regularizers.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+
+class L2(object):
```
hussainsultan: In general, i am not a big fan of classes with just static methods. If name spacing is important, we should pull them out into separate files/modules. Alternatively, we could use a factory class that configures these methods.

dask_glm/tests/test_algos_families.py
```diff
@@ -80,24 +81,25 @@ def test_basic_unreg_descent(func, kwargs, N, nchunks, family):
...
     (admm, {'abstol': 1e-4}),
-    (proximal_grad, {'tol': 1e-7, 'reg': 'l1'}),
+    (proximal_grad, {'tol': 1e-7}),
 ])
 @pytest.mark.parametrize('N', [1000])
 @pytest.mark.parametrize('nchunks', [1, 10])
 @pytest.mark.parametrize('family', [Logistic, Normal])
 @pytest.mark.parametrize('lam', [0.01, 1.2, 4.05])
-def test_basic_reg_descent(func, kwargs, N, nchunks, family, lam):
+@pytest.mark.parametrize('reg', [L1, L2])
```
hussainsultan: These tests are really nice and super readable!

dask_glm/algorithms.py
moody-marlin: Nope, it works with *any* regularizer which is one reason it's awesome!  If you look at pg. 65 of the [original paper](http://web.stanford.edu/~boyd/papers/pdf/admm_distr_stats.pdf), the `z` update is just computing the "proximal operator" for the regularizer.  In the case of `l1`, that's shrinkage.

dask_glm/algorithms.py
moody-marlin: Hm yea, it's not super obvious; this is creating what you used to call the "proximal_logistic_loss" -- it's the local function that is being optimized on each chunk.

dask_glm/regularizers.py
moody-marlin: Yea, that makes sense; I don't have a strong opinion on this, other than to say the current organization is easy (for me personally) to manage.

PR #36, opened 2017-03-21
Add tests for deterministic output

mrocklin:


mrocklin:
I tried and failed to reproduce an in-place failure with threads:

### Sequential

```python
import numpy as np
np.__config__.show()
import numpy as np
x = np.random.random((2048, 2048))
results = [x.dot(x) for i in range(10)]
np.array(results).all(axis=-1).all()
# True
```

### Parallel

```python
from concurrent.futures import ThreadPoolExecutor
e = ThreadPoolExecutor(10)
for i in range(20): e.submit(lambda x: x + 1, i)  # warm up threads
%time futures = [e.submit(lambda: x.dot(x)) for _ in range(20)]
%time futures2 = [e.submit(lambda x, y: np.log(x) + y, x, i) for i in range(20)]
results = [f.result() for f in futures]
results2 = [f.result() for f in futures2]
np.array(results).all(axis=-1).all() and np.array(results2).all(axis=-1).all()
# True
```

moody-marlin:
LGTM, thanks!  I also have not been able to reproduce a failure.

PR #37, opened 2017-03-21
Small adjustments to ADMM

### LINKED ISSUE #12: ADMM: Inner L-BFGS optimizations fail to converge
If I slightly modify the inner `local_update` solves in the current ADMM implementation as follows:

```
def local_update(y, X, w, z, u, rho, fprime=proximal_logistic_gradient,
                 f=proximal_logistic_loss,
                 solver=fmin_l_bfgs_b):
    w = w.ravel()
    u = u.ravel()
    z = z.ravel()
    solver_args = (X, y, z, u, rho)
    w, f, d = solver(f, w, fprime=fprime, args=solver_args, pgtol=1e-10,
                     maxiter=200,
                     maxfun=250, factr=1e-30)
    if d['warnflag']:
        raise ValueError(
            '''Internal LBFGSB algorithm failed to converge!
            Details: {}'''.format(d['task']))
    else:
        return w.reshape(2, 1)
```

Then on random data I see convergence issues:

```
from dask_glm.utils import make_y
X = da.random.random((100,2), chunks=(100/5, 2))
y = make_y(X)
logistic_regression(X, y[:,None], 0.1, 0.1, 1)
```

With the following error:
```
Traceback (most recent call last):
...
dask.async.ValueError: Internal LBFGSB algorithm failed to converge!
            Details: b'ABNORMAL_TERMINATION_IN_LNSRCH'
```
which refers to abnormal termination in the line search (always with the line search...).

From the `fmin_l_bfgs_b` output:
```
Line search cannot locate an adequate point after 20 function
  and gradient evaluations.  Previous x, f and g restored.
 Possible causes: 1 error in function or gradient evaluation;
                  2 rounding error dominate computation.
```

moody-marlin:
- Update the convergence checks to have proper scaling (depending on chunk-size).
- Closes #12 by changing the stringent defaults in the `scipy.optimize` call

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -136,7 +136,7 @@ def newton(X, y, max_steps=50, tol=1e-8, family=Logistic):
 
 
 def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
-         max_steps=100, abstol=1e-4, reltol=1e-2, family=Logistic):
+         max_steps=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
@@ -180,7 +180,6 @@ def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
         zold = z.copy()
         ztilde = np.mean(beta_hat + np.array(u), axis=0)
         z = reg.proximal_operator(ztilde, lamduh / (rho * nchunks))
-#        z = shrinkage(ztilde, lamduh / (rho * nchunks))
 
         # u-update step
         u += beta_hat - z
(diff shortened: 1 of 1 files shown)


hussainsultan:
lgtm

mrocklin:
I merged this but perhaps I shouldn't have.  It makes our recent benchmark perform worse.  What used to take 30s now takes minutes.  That benchmark is pretty artificial though.

moody-marlin:
@mrocklin In terms of output quality, is it the same, worse or better?  Ultimately this only changes the *default* tolerance settings to something that scales appropriately with increasing chunk-sizes -- that change does in fact make the convergence criterion a little bit tighter, so I would expect more iterations.  You could always achieve the old convergence times by adjusting the `reltol` parameter down, I'm guessing something like `reltol=5e-1` would work.

Issue #30, opened 2017-02-24
BFGS: to keep or not to keep?


moody-marlin:
Currently, the implementation of `bfgs` isn't always converging, and the implementation is messy.  Should we spend some time working on this, or toss it?  If we're going to keep it, what exactly is going on with the implementation?

MLnick:
By the way is it possible to just use `scipy.optimize`? That way it will certainly be correct?

Not sure if you are using standard L-BFGS (with parallel gradient computation, sum the worker grads on master, take L-BFGS step) or the vector-free version (in which case I guess you would need to do your own impl)

mrocklin:
I believe that scipy.optimize will convert things to numpy arrays.  These may not fit in local memory.  The current implementation keeps everything on the cluster except the final parameters.

stoneyv:
The **scipy.optimize L-BFGS-B optimization method** passes the arrays to a Fortran routine via F2PY.  
https://github.com/scipy/scipy/blob/master/scipy/optimize/lbfgsb/lbfgsb.pyf  

scipy.optimize offers numerous optimization methods including L-BFGS-B. (L limited memory, B with simple bounds ) The L-BFGS-B implementation utilizes Fortran code from Richard H. Byrd et Al at NorthWestern university via F2PY which became part of numpy in 2007.  
https://github.com/scipy/scipy/tree/master/scipy/optimize/lbfgsb  

I see two candidates for BFGS when dask is distributed. This may also allow for a single node implementation that does not use as much memory as the Northwester university Fortran L-BFGS-B implementation.  Xiaocheng Tang implemented a distributed  L-BFGS in scala using (2) to replace the non distributed MLLib BFGS
Xiaocheng Tang's distributed L-BFGS Apache Spark implementation from (2)
https://github.com/LHAC/dac  
MLLib optimization LBFGS.scala
https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/optimization/LBFGS.scala

1. **Decentralized Quasi-Newton Methods** 
decentralized Broyden-Fletcher-Goldfarb-Shanno D-BFGS as variation of BFGS quasi-Newton  
"fully distributed where nodes approximate curvature info of self and neighbor by satisfaction of secant condition.  "
https://arxiv.org/pdf/1605.00933.pdf

2. **Large-scale L-BFGS using MapReduce**
Weizhu Chen, Zhenghao Wang, Jingren Zhou
Microsoft
[...]

PR #40, opened 2017-04-12
API: Sketch out Estimator API

TomAugspurger:
Added a couple Estimators for an end-user API.

Basically

```python
from dask_glm.estimators import LogisticRegression
lr = LogisticRegresssion()
lr.fit(X, y)
```

To kick the [API discussion](https://github.com/dask/dask-glm/issues/11) even further down the road, right now it accepts either scikit-learn style `Estimator(...).fit(X, y)` or statsmodels-style `Estimator(X, y, ...).fit()`, but we should probably choose one or the other.

Here's an [example notebook](http://nbviewer.jupyter.org/gist/anonymous/ceed00b4c0518f88db71dabefc3afeff)

Some miscellaneous changes to help

- Added a `datasets.make_classification` helper
- Added string names for solvers
- Added an `add_intercept` helper ~(which I think might be broken right now...)~

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -1,3 +1,18 @@
+"""
+
+Parameter Key:
+
+================  =========  ===  ======  ===========  =======  ===  ==========  ======  ======
+algo / parameter  max_iter   tol  family  regularizer  lambduh  rho  over_relax  abstol  reltol
+================  =========  ===  ======  ===========  =======  ===  ==========  ======  ======
+admm              X          *    X       X            X        X    X           X       x
+gradient_descent  X          X    X       .            .        .    .           .       .
+newton            X          X    X       .            .        .    .           .       .
+bfgs              X          X    X       .            .        .    .           .       .
+proximal_grad     X          X    X       X            X        .    .           .       .
+================  =========  ===  ======  ===========  =======  ===  ==========  ======  ======
+"""
+
 from __future__ import absolute_import, division, print_function
 
 from dask import delayed, persist, compute
@@ -9,7 +24,7 @@ from scipy.optimize import fmin_l_bfgs_b
 
 from dask_glm.utils import dot, exp, log1p
 from dask_glm.families import Logistic
-from dask_glm.regularizers import L1
+from dask_glm.regularizers import L1, _regularizers
 
 
 def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
@@ -41,7 +56,7 @@ def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
     return stepSize, beta, Xbeta, func
 
 
-def gradient_descent(X, y, max_steps=100, tol=1e-14, family=Logistic):
+def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic):
     '''Michael Grant's implementation of Gradient Descent.'''
 
     loglike, gradient = family.loglike, family.gradient
@@ -55,7 +70,7 @@ def gradient_descent(X, y, max_steps=100, tol=1e-14, family=Logistic):
     backtrackMult = firstBacktrackMult
     beta = np.zeros(p)
 
-    for k in range(max_steps):
+    for k in range(max_iter):
         # how necessary is this recalculation?
         if k % recalcRate == 0:
             Xbeta = X.dot(beta)
@@ -97,7 +112,7 @@ def gradient_descent(X, y, max_steps=100, tol=1e-14, family=Logistic):
     return beta
 
 
-def newton(X, y, max_steps=50, tol=1e-8, family=Logistic):
+def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
     '''Newtons Method for Logistic Regression.'''
 
     gradient, hessian = family.gradient, family.hessian
@@ -127,7 +142,7 @@ def newton(X, y, max_steps=50, tol=1e-8, family=Logistic):
         # should change this criterion
         coef_change = np.absolute(beta_old - beta)
         converged = (
-            (not np.any(coef_change > tol)) or (iter_count > max_steps))
+            (not np.any(coef_change > tol)) or (iter_count > max_iter))
 
         if not converged:
             Xbeta = dot(X, beta)  # numpy -> dask converstion of beta
@@ -135,11 +150,12 @@ def newton(X, y, max_steps=50, tol=1e-8, family=Logistic):
     return beta
 
 
-def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
-         max_steps=250, abstol=1e-4, reltol=1e-2, family=Logistic):
+def admm(X, y, regularizer=L1, lamduh=0.1, rho=1, over_relax=1,
+         max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
+    regularizer = _regularizers.get(regularizer, regularizer)  # string
 
     def create_local_gradient(func):
         @functools.wraps(func)
@@ -157,16 +173,25 @@ def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
     f = create_local_f(pointwise_loss)
     fprime = create_local_gradient(pointwise_gradient)
 
-    nchunks = X.npartitions
+    nchunks = getattr(X, 'npartitions', 1)
+    # nchunks = X.npartitions
     (n, p) = X.shape
-    XD = X.to_delayed().flatten().tolist()
-    yD = y.to_delayed().flatten().tolist()
+    # XD = X.to_delayed().flatten().tolist()
+    # yD = y.to_delayed().flatten().tolist()
+    if isinstance(X, da.Array):
+        XD = X.rechunk((None, X.shape[-1])).to_delayed().flatten().tolist()
+    else:
+        XD = [X]
+    if isinstance(y, da.Array):
+        yD = y.rechunk((None, y.shape[-1])).to_delayed().flatten().tolist()
+    else:
+        yD = [y]
 
     z = np.zeros(p)
     u = np.array([np.zeros(p) for i in range(nchunks)])
-    betas = np.array([np.zeros(p) for i in range(nchunks)])
+    betas = np.array([np.ones(p) for i in range(nchunks)])
 
-    for k in range(max_steps):
+    for k in range(max_iter):
 
         # x-update step
         new_betas = [delayed(local_update)(xx, yy, bb, z, uu, rho, f=f,
@@ -179,7 +204,7 @@ def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
         #  z-update step
         zold = z.copy()
         ztilde = np.mean(beta_hat + np.array(u), axis=0)
-        z = reg.proximal_operator(ztilde, lamduh / (rho * nchunks))
+        z = regularizer.proximal_operator(ztilde, lamduh / (rho * nchunks))
 
         # u-update step
         u += beta_hat - z
@@ -218,7 +243,7 @@ def shrinkage(x, kappa):
     return z
 
 
-def bfgs(X, y, max_steps=500, tol=1e-14, family=Logistic):
+def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
     '''Simple implementation of BFGS.'''
 
     n, p = X.shape
@@ -232,7 +257,7 @@ def bfgs(X, y, max_steps=500, tol=1e-14, family=Logistic):
 
     beta = np.zeros(p)
     Hk = np.eye(p)
-    for k in range(max_steps):
+    for k in range(max_iter):
 
         if k % recalcRate == 0:
             Xbeta = X.dot(beta)
@@ -294,8 +319,8 @@ def bfgs(X, y, max_steps=500, tol=1e-14, family=Logistic):
     return beta
 
 
-def proximal_grad(X, y, reg=L1, lamduh=0.1, family=Logistic,
-                  max_steps=100, tol=1e-8, verbose=False):
+def proximal_grad(X, y, regularizer=L1, lamduh=0.1, family=Logistic,
+                  max_iter=100, tol=1e-8, verbose=False):
 
     n, p = X.shape
     firstBacktrackMult = 0.1
@@ -306,12 +331,13 @@ def proximal_grad(X, y, reg=L1, lamduh=0.1, family=Logistic,
     recalcRate = 10
     backtrackMult = firstBacktrackMult
     beta = np.zeros(p)
+    regularizer = _regularizers.get(regularizer, regularizer)  # string
 
     if verbose:
         print('#       -f        |df/f|    |dx/x|    step')
         print('----------------------------------------------')
 
-    for k in range(max_steps):
+    for k in range(max_iter):
         # Compute the gradient
         if k % recalcRate == 0:
             Xbeta = X.dot(beta)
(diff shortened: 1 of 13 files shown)


dask_glm/estimators.py
```diff
@@ -0,0 +1,144 @@
...
+        self.fit_intercept = True
+        self.solver = algorithms._solvers[solver]
+        if solver in ('admm', 'proximal_grad'):
+            regularizer = regularizer or 'L1'
+            lamduh = lamduh if lamduh is not None else 0.1
+        if solver == 'admm':
+            rho = rho if rho is not None else 1
+            over_relax = over_relax if over_relax is not None else 1
+            abstol = abstol if abstol is not None else 1e-4
+            reltol = reltol if reltol is not None else 1e-2
```
mrocklin: I think that we should avoid calling out algorithms explicitly by name here if possible.  Is it possible to put these defaults in algorithms.py somehow?

dask_glm/estimators.py
```diff
@@ -0,0 +1,144 @@
...
+        self.abstol = abstol
+        self.reltol = reltol
+        self._validate()
+        self.coef_ = None
+        self.intercept_ = None
+        self._coef = None  # coef, maybe with intercept
+
+        fit_kwargs = {'max_steps', 'tol', 'family'}
+        # TODO: use scikit-learn names throughout the lib
+        renamer = {'max_steps': 'max_iter', 'reg': 'regularizer'}
```
mrocklin: I think it would also be reasonable to change the existing algorithms to use these names consistently.  I wouldn't be shy about touching that code.

dask_glm/estimators.py
```diff
@@ -0,0 +1,137 @@
...
+Models following scikit-learn's estimator API.
+"""
+from sklearn.base import BaseEstimator
+
+from . import algorithms
+from . import families
+from .utils import sigmoid, dot, add_intercept
+
+
+class _GLM(BaseEstimator):
```
moody-marlin: +1 for `_GLM` being a base-class.

dask_glm/estimators.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def predict(self, X):
+        X_ = self._maybe_add_intercept(X)
+        return dot(X_, self._coef)
+
+    def score(self, X, y_true):
+        return mean_square_error(y, self.predict(X))
+
+
+def accuracy_score(y_true, y_pred):
```
moody-marlin: Should these functions be in:
- `utils` or
-  within the corresponding Estimator class
?

TomAugspurger:
Here's an update to the notebook showing scikit-learn and dask on a slice of the taxi dataset

http://nbviewer.jupyter.org/gist/anonymous/c07a617684ec9629c204825c212383fe

It shows dask-glm interacting with scikit-learn's pipeline API, and a small bit at the end with dask-searchcv. I haven't looked closely, at the output, but the fit doesn't seem great. Not sure if it's the defaults / starting parameters or something else.

dask_glm/algorithms.py
```diff
@@ -127,19 +142,20 @@ def newton(X, y, max_steps=50, tol=1e-8, family=Logistic):
...
 
 
-def admm(X, y, reg=L1, lamduh=0.1, rho=1, over_relax=1,
-         max_steps=250, abstol=1e-4, reltol=1e-2, family=Logistic):
+def admm(X, y, regularizer=L1, lamduh=0.1, rho=1, over_relax=1,
+         max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
+    regularizer = _regularizers.get(regularizer, regularizer)  # string
```
moody-marlin: Ha this line took me a second.

dask_glm/estimators.py
TomAugspurger: Definitely not at the bottom of `estimators.py`, forgot to move them before pushing :) I didn't want to go down the whole rabbit-hole of our own `metrics.py`, but maybe we should do a couple basic ones like this.

dask_glm/estimators.py
```diff
@@ -0,0 +1,137 @@
...
+            fit_kwargs.update({
+                'regularizer', 'lamduh', 'rho', 'over_relax', 'abstol',
+                'reltol'
+            })
+        elif solver == 'proximal_grad':
+            fit_kwargs.update({'regularizer', 'lamduh'})
+
+        self._fit_kwargs = {k: getattr(self, k) for k in fit_kwargs}
+
+    def fit(self, X, y=None):
```
moody-marlin: Whenever we include `p`-value output, we will need access to the data that was used to *fit* the model (in order to compute the Hessian).  

We could have this post-fit processing inside the `fit` method with an inert call to `self.post_process` or something as a place-holder, or store that info in some other cleverly-named attributes?

mrocklin:
Cool stuff.  Comments on the notebook:

1.  da.asarray may not operate as expected on dask.dataframes.  You might want to try `df.values` instead
2.  for `overflow = overflow.compute() if hasattr(overflow, 'compute') else overflow` you might want to try `dask.compute(overflow)[0]` instead.  This function was designed to pass through non-dask things harmlessly
3.  "Converged".  Looks like we need to remove print statements from dask-glm
4.  What do you think we should do with things like `DummyEncoder`?

moody-marlin:
Comments on the notebook:
- the model fit by `scikit-learn` isn't immediately comparable to the model fit by `dask-glm` because of regularization strength.  To make them more comparable, do:
``` 
lam = 0.1 # or whatever number you want <-- reg strength for dask-glm
C = 1 / lam # <--- reg strength for scikit-learn
```

- I'd recommend using `admm` as the default solver for regularized problems
- +1 on removing print statements; I can do a separate PR that stores whether the chosen solver actually converged so the user has access to that.

TomAugspurger:
> What do you think we should do with things like DummyEncoder?

I took it from https://github.com/TomAugspurger/sktransformers/blob/master/sktransformers/preprocessing.py

I'm not sure of the best home for them. I don't think we can reasonably expect the scikit-learn devs to "do the right thing" for dask stuff, like they're mostly doing for pandas dataframes.

That said, it's not a *ton* of work to support regular- and dask-versions of these kinds of tranformers. The `DummyEncoder` one is so long because there isn't anything exactly like it in scikit-learn. I could probably take on building-out / maintaining this if needed.

moody-marlin:
Oh, an additional comment: it looks like `scikit-learn` also defaults to `l2` regularization, whereas I think `dask-glm` defaults to `l1`.  (writing this out also makes me wonder if we should lower-case these to better fit the mathematical notation).

We might want to set all of our defaults to `scikit-learn` defaults to avoid confusion whenever people run these sorts of comparison experiments in the future.

dask_glm/estimators.py
moody-marlin: I'd be comfortable including these in `utils.py` for now.

dask_glm/estimators.py
moody-marlin: The only other call-out I'll make here is that this initialization setup will make it difficult for a user to create an unregularized model (they will need to specify a hand-picked solver).  I'm not worried about it for this PR, but I think we should be aware of it and make it easier (without having the user choose the solver) in the future.

moody-marlin:
If you could move those stray functions around, change the default regularizer to 'l2' and the default `lamduh` to 1.0 (to match `sci-kit`) then I think we're good to go.

TomAugspurger:
Cool, I'll get to those this afternoon.

TomAugspurger:
Ok, rebased  and addressed those last couple comments @moody-marlin 

And for fun, another example on the [SUSY dataset](https://archive.ics.uci.edu/ml/datasets/SUSY), (5,000,000 x 18; all numeric): http://nbviewer.jupyter.org/gist/anonymous/15742155693794ddd31ea85b654cbc7e

The takeaway, assuming I've done everything properly

| Library      | Training time | Score |
| -------------| ------------- | ----- |
| dask-glm     | 9:25          | .788  |
| scikit-learn | 25:01         | .788  |

That's all on my local machine still.

moody-marlin:
Awesome, LGTM after flaking.

mrocklin:
It's very cool to see those performance numbers.  cc @amueller @ogrisel in case they're interested.  Notebook URL copied from above for convenience:

> And for fun, another example on the SUSY dataset, (5,000,000 x 18; all numeric): http://nbviewer.jupyter.org/gist/anonymous/15742155693794ddd31ea85b654cbc7e

amueller:
Nice. how many cores was that?

TomAugspurger:
I believe it was 8 cores.

mrocklin:
8 logical or 8 physical cores.  My original assumption was that you were doing this from a personal laptop, presumably running something like a core i7 which has four physical cores with hyperthreading up to eight logical cores.

TomAugspurger:
Ah, 8 logical cores I suppose. I was just going off the number of workers
that `Client` created (8). Looking in the settings there are indeed 4 cores.

ogrisel:
Nice though I get `0.78832579999999997` accuracy in ~1min with a single core and the 'saga' solver in sklearn master (on standardized data) ;)

mrocklin:
Heh, I wonder how large of a cluster we would need to beat @ogrisel 's single laptop core :)

amueller:
you need a dataset that doesn't fit into his ram ;)

mrocklin:
@TomAugspurger @moody-marlin if you were interested, I think that this would be an entertaining blogpost to put out there:

"Dask-glm beats scikit-learn by using parallelism.  Scikit-learn then beats dask-glm by using smarter algorithms" :)

It would both show off the sklearn API compatibility developed in this PR and would also have the larger message of "there are usually better options than parallel computing".

ogrisel:
Also I checked the resulting `coef_` attribute and no coefficient is set to zero for this value of `C`. The l1 penalty probably does not make sense on such low dimensional data. I suspect XGBoost to perform much better on this kind of data.

ogrisel:
To dask-glm's defense, the fit takes only 6min on 2 cores when you scale the data (same final training accuracy). saga is still 6x faster on a single core though.

Scaling the data is always a good idea :) and is required for the saga solver otherwise it might not converge.

mrocklin:
More broadly, are there parallelizable convex optimization algorithms that we should implement in dask-glm that would likely perform better on this sort of problem than what is being used now (ADMM I think, for this problem).

ogrisel:
I am not very familiar with ADMM and its competitors. Stochastic sequential solvers are very efficient for gradient based estimators when the number of samples is large.

Still @fabianp presented last week a parallel version of SAGA at AISTATS:

https://twitter.com/fpedregosa/status/856565126744334336

ogrisel:
For reference here is my notebooks (evaluated on my laptop with 2 physical cores):

http://nbviewer.jupyter.org/gist/ogrisel/5f2d31bc5e7df852b4ca63f5f6049f42

mrocklin:
@fabianp is this the right paper to read: https://arxiv.org/pdf/1606.04809.pdf ?

ogrisel:
@TomAugspurger if you do a blog post, you should use the official train test split (last 500000 samples as held out test set) and report accuracy on both train set (to assess the quality of the optimizers) and test set (to assess the quality of the statistical model).

TomAugspurger:
Thanks, will do.

moody-marlin:
I'd like to throw in a quick comment: we shouldn't be trying to beat any benchmarks for output quality / predictive power; `dask-glm` is just fitting a simple regularized logistic regression, and so is scikit-learn.  We just want our output to be the true optimizer for the problem the user has provided.  I believe reporting the `score` metrics was initially being used as a proxy to prove our two algorithms were converging on the same "true" MAP coefficient estimates (i.e., our output is just as reliable as `scikit-learn`).

I believe the `sag` / `saga` algorithm in `scikit-learn` only fits differentiable regularizers, i.e., it only works with `l2` regularization, so comparing it against `l1` regularized fits isn't a comparison of our packages' performance but merely a statement that for this dataset `l2` regularization works better than `l1` (and is faster, which is typically true because of differentiability).

ogrisel:
SAGA works well with l1.

moody-marlin:
Ah you're right, it works with the proximal operator instead of the gradient. :+1:

fabianp:
@ogrisel: thanks for the plug ;-)

@mrocklin: this is indeed the right paper (https://arxiv.org/pdf/1606.04809.pdf)

While this asynchronous algorithm works really well in practice, I highlight a couple of implementation caveats just to be 100% clear on it. I am not familiar with Dask, so feel free to comment, I love to have feedback on this:

   * The algorithm is fully asynchronous except for the per-coefficient update, which is assumed to be atomic (lines 11-13 in Algorithm 2), which is the common setting for these algorithms.  That is, there is no locks on the full vector of coefficients as in other synchronous algorithms, but the individual scalar updates to the vector of coefficient do need to be atomic. This means that in order to get convergence to machine precision (one can argue whether this is necessary from a machine learning perspective, but it is important from an optimization perspective), one needs to have atomic double operations (we also refer to these as compare-and-swap operations in the paper). This turns out to be necessary not only in theory, but also in pratice, see Fig. 6 and paragraph "Necessity of compare-and-swap operations".
I could do this in Java and C++, but I couldn't find a way to do this (efficiently) in Python. 

   * The asynchronous algorithm in that paper is restricted to smooth objectives, but it can be extended to proximal terms. We have a paper under review on this that we will post in the next weeks on ArXiv.

(apologies for side-tracking the conversation and thanks for dask-glm!)

mrocklin:
Yeah, Dask certainly isn't going to operate at the low latencies required by algorithms such as you've implemented.  Dask today imposes latencies on the order of low tens of milliseconds.  This is almost certainly well above what your algorithm can tolerate.

However, my understanding is that these same sorts of algorithms sometimes develop blocked variants that are more tolerant of high latencies and so more amenable to distributed parallelism.  Is this the case for asynchronous SAGA?

ogrisel:
In practice in that case it would probably make sense to run `n` instances of the penalized logistic regression fit with different values of C and on different CV splits to find the best model (in terms of validation accuracy). This is embarrassingly parallel so would well suited for `dask-glm` (even if less interesting that distributed optimizers from an implementer's point of view).

In this is basically what is implemented by `LogisticRegressionCV` in scikit-learn (using single machine threads via joblib).

ogrisel:
Out of core SAGA would be nice too.

ogrisel:
> However, my understanding is that these same sorts of algorithms sometimes develop blocked variants that are more tolerant of high latencies and so more amenable to distributed parallelism. Is this the case for asynchronous SAGA?

Blocked Coordinate Descent is competitive for L1 (or L1+L2) penalized LogisticRegression when n_features >> n_samples. But naive implementations are bad compared to variants with screening rules and working sets. As soon as you implement those, parallelism might get trickier (or even impossible) but I am not expert enough to tell. @agramfort or @mblondel might know better.

agramfort:
before doing something heavy did you consider a distributed L-BFGS where
you distribute the gradient computations? You can use L-BFGS-B for L1
penalty with a tiny trick.

mrocklin:
@MLnick did a quick version of this with decent results over here: https://github.com/dask/dask-glm/issues/30#issuecomment-292155158

fabianp:
I’m afraid that block (or batch) versions of asynchronous SGD or SAGA are less attractive than for coordinate descent. The reason is that being able to do sparse updates (in the sense of modifying a small subset of coefficients at each iteration) is a big part of the efficiency of asynchronous SGD/SAGA, and batching the updates would destroy the sparsity in the updates.

MLnick:
@agramfort interesting for L1 with L-BFGS-B. I thought the OWLQN variant was required for L1. Do you mind sharing what the "trick" is? :)

On the subject of distributed SGD - I'm quite keen to look into a "parameter server" type version of parallel SGD on dask - so either fully async or bounded stale sync. With a distributed parameter vector. I _think_ that it may be quite nice to do it via the local `Client` interface (or perhaps `Channel`?) but I haven't had a chance to really dig into it just yet.

For large sparse problems I think it would work well (e.g. sparse LR and FMs cf. DiFacto).

agramfort:
see:

https://gist.github.com/vene/fab06038a00309569cdd

you basically split the coefs in positive and negative part.

PR #42, opened 2017-04-25
Support sparse arrays

mrocklin:
I've been playing recently with sparse arrays in https://github.com/mrocklin/sparse and in scipy.sparse.  I had to generalize the dispatched functions in utils.py a bit.  It would be nice to eventually depend on `__numpy_ufunc__` for most of this in the future.


moody-marlin:
LGTM -- I'm saying this off-the-cuff, but I wonder if we could work out an algorithm that can efficiently handle data which has a dense and a sparse component, similar to what we saw on the taxi cab dataset.

mrocklin:
Yeah, I think that we'll probably be able to make dask.array operations handle that decently well.  I don't know how best to handle this with delayed computations like in admm, except that maybe they too operate on two-chunk dask.arrays.

PR #43, opened 2017-04-29
Sparse+dense mixed arrays

mrocklin:
Previously admm would rechunk the columns to be in a single chunk, and
then pass delayed numpy arrays to the local_update function.  If the
chunks along columns were of different types, like a numpy array and a
sparse array, then these would be inefficiently coerced to a single
type.

Now we pass a list of numpy arrays to the local_update function.  If
this list has more than one element then we construct a local dask.array
so that operations like dot do the right thing and call two different
local dot functions, one for each type.

This currently depends on https://github.com/dask/dask/pull/2272 though I may be able to avoid this dependency.

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -185,8 +185,6 @@ def admm(X, y, regularizer=L1, lamduh=0.1, rho=1, over_relax=1,
 
     if isinstance(X, da.Array):
         XD = X.to_delayed().tolist()
-        if len(XD[0]) == 1:
-            XD = [x[0] for x in XD]
     else:
         XD = [X]
 
@@ -237,10 +235,12 @@ def maybe_compute(x):
 
 
 def local_update(X, y, beta, z, u, rho, f, fprime, solver=fmin_l_bfgs_b):
-    if isinstance(X, list):
+    if len(X) > 1:
         X = [da.from_array(x, chunks=x.shape, name=False,
                            getitem=operator.getitem) for x in X]
         X = da.concatenate(X, axis=1)
+    else:
+        X = X[0]
 
     f2 = toolz.compose(maybe_compute, f)
     fprime2 = toolz.compose(maybe_compute, fprime)
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -15,12 +15,15 @@ proximal_grad     X          X    X       X            X        .    .          
 
 from __future__ import absolute_import, division, print_function
 
-from dask import delayed, persist, compute
 import functools
-import numpy as np
-import dask.array as da
-from scipy.optimize import fmin_l_bfgs_b
+import operator
 
+import dask
+from dask import delayed, persist, compute
+import dask.array as da
+import numpy as np
+from scipy.optimize import fmin_l_bfgs_b
+import toolz
 
 from dask_glm.utils import dot, exp, log1p
 from dask_glm.families import Logistic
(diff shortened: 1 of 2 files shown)


mrocklin:
There is a non-trivial cost to using dask.array within the function given to `scipy.optimize`.  Graph generation costs appear to be non-trivial.  I can reduce these somewhat, but I'm also curious if it is possible to apply `f` and `fprime` individually to all chunks of the input to `local_update`.  In this case each chunk correspond to a block of *columns*.  Looking at the methods in families.py it looks like it might be possible to evaluate `f` on each block and add them together and to evaluate `fprime` on each block and concatenate them.  

@moody-marlin is this generally possible?

(please let me know if this description was not clear)

moody-marlin:
Hmmm I'm not sure how possible this is; it definitely won't be as easy as evaluating `f` and `fprime` on each block and combining them though; for example, 
```
(( y - X1.dot(beta) - X2.dot(beta)) ** 2).sum()
```
doesn't split as a simple sum on each chunk.

There might be fancier ways of combining the results, or possibly even altering ADMM to take this into account, but it will require some non-trivial thinking.

mrocklin:
> There might be fancier ways of combining the results

The fancy way here is already handled by dask.array.  I was just hoping to avoid having to recreate graphs every time.  I can probably be clever here though.  I'll give it a shot

mrocklin:
OK, I've pushed a solution that, I think, avoids most graph construction costs.  However my algorithm is failing to converge.  @moody-marlin if you find yourself with a free 15 minutes can you take a look at `def local_update()` and see if I'm doing anything wonky?  I suspect that there is something dumb going on.  When I print out the result and gradients I find that it doesn't seem to converge (gradients stay large).

dask_glm/algorithms.py
```diff
@@ -219,21 +224,47 @@ def wrapped(beta, X, y, z, u, rho):
...
+    else:
+        X = X[0]
+        solver_args = (X, y, z, u, rho)
+
+        def f2(beta, *args):
+            result, gradient = f(beta, *args), fprime(beta, *args)
+            print(result, gradient)
+            return result, gradient
+
+    beta, _, _ = solver(f2, beta, args=solver_args, maxiter=200, maxfun=250)
```
moody-marlin: ^^^ I think this line is incorrect; solver (in this case `fmin_l_bfgs_b`) has [this](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.optimize.fmin_l_bfgs_b.html) call signature; the first argument is the function `f`, and a *keyword* argument needs to be `fprime`.

mrocklin: I thought that I didn't have to specify fprime if f returned two results

moody-marlin: Oh, it looks like you're right; sorry, I've never called it that way. Hmm back to the drawing board.

dask_glm/algorithms.py
```diff
@@ -219,21 +224,47 @@ def wrapped(beta, X, y, z, u, rho):
...
+
+        beta_key = beta._keys()[0]
+        def f2(beta):
+            """ Reuse existing graph, just swap in new beta and compute """
+            dsk[beta_key] = beta
+            result, gradient = dask.get(dsk, [ff.key, ffprime.key])
+            print(result, gradient)
+            return result, gradient
+
+        solver_args = ()
```
moody-marlin: Why no `solver_args` in this case?

mrocklin: They are all, I think, in the task graph.  My assumption is that these will not change during the call to local_update.  Is this correct?

moody-marlin: Yea that's correct.

PR #44, opened 2017-05-01
Logistic overflow and clean-up

### LINKED ISSUE #41: ADMM is susceptible to overflow
(edited to not require the new API)

Probably other algorithms too. I see that proximal_grad does handle it correctly.
This little script compares the fit for scikit-learn LogisticRegression and our admm against a baseline of no-overflow, to a copy of the dataset with `n` values in a range that will overflow when passed through the sigmoid function.

```python
import warnings

import dask.array as da
import numpy as np
from scipy.optimize import fmin_l_bfgs_b
from sklearn import datasets
from sklearn.linear_model import LogisticRegression
from dask_glm.algorithms import admm

warnings.simplefilter("ignore", RuntimeWarning)

C = 10
lambda_ = 1 / C
np.random.seed(2)
X, y = datasets.make_classification(n_classes=2, n_samples=1000)

Xda = da.from_array(X, (10, X.shape[1]))
yda = da.from_array(y, (10,))

lrbase = LogisticRegression(fit_intercept=False, C=C, max_iter=10).fit(X, y)
admm_base = admm(Xda, yda, lamduh=lambda_, max_iter=10)


def fit_and_compare(X, y, n):
    X = X.copy()
    np.random.seed(2)
    idx = np.random.choice(np.arange(len(X)), size=(n,))
    X[idx, 1] = np.random.randint(750, 1000, size=(n,))

    lr = LogisticRegression(fit_intercept=False, C=C, max_iter=10)
    lr.fit(X, y)

    Xda = da.from_array(X, (10, X.shape[1]))
    yda = da.from_array(y, (10,))
    result = admm(Xda, yda, lamduh=lambda_, max_iter=10)

    print(f"Overflow={n}")
    print("Sklearn:", np.mean(np.abs(lrbase.coef_ - lr.coef_)))
    print("admm   :", np.mean(np.abs(admm_base - result)))
    return lr.coef_, result


fit_and_compare(X, y, 0)
fit_and_compare(X, y, 5)
fit_and_compare(X, y, 10)
fit_and_compare(X, y, 50)
fit_and_compare(X, y, 100)
```

Outputs

```
Overflow=0
Sklearn: 6.21681525703e-17
admm   : 1.04008099876e-13
Overflow=5
Sklearn: 0.00377521192273
admm   : 0.0975875697572
Overflow=10
Sklearn: 0.00311021598129
admm   : 0.118998783966
Overflow=50
Sklearn: 0.00055956264663
admm   : 0.178748360839
Overflow=100
Sklearn: 0.00312859463805
admm   : 0.196356993923
```

moody-marlin:
An initial attempt to address #41:

- [x] reduce overflows in calls to `Logistic.loglike`

- [x] remove all `print` statements

- [x] add test for `numpy` output

- [x] fix failing distributed determinism test

- [x] determine how to handle scaling in the absence of an intercept

This *doesn't* close #41 yet, because I don't currently scale inputs when an intercept is not present to absorb the means.

I'm not sure why I made `normalize` a decorator, it seemed like the right thing to do at the time; alternatively, we could just make it a keyword argument to both the estimators and the algorithms.  Open to input here.

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -150,6 +150,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     return beta
 
 
+@normalize()
 def admm(X, y, regularizer='l1', lamduh=0.1, rho=1, over_relax=1,
          max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
@@ -237,11 +238,6 @@ def local_update(X, y, beta, z, u, rho, f, fprime, solver=fmin_l_bfgs_b):
     return beta
 
 
-def shrinkage(x, kappa):
-    z = np.maximum(0, x - kappa) - np.maximum(0, -x - kappa)
-    return z
-
-
 @normalize()
 def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic, **kwargs):
     '''Simple implementation of BFGS.'''
@@ -316,6 +312,7 @@ def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic, **kwargs):
     return beta
 
 
+@normalize()
 def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
                   max_iter=100, tol=1e-8):
 
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -22,7 +22,7 @@ import dask.array as da
 from scipy.optimize import fmin_l_bfgs_b
 
 
-from dask_glm.utils import dot, exp, log1p
+from dask_glm.utils import dot, exp, log1p, normalize
 from dask_glm.families import Logistic
 from dask_glm.regularizers import L1, _regularizers
 
@@ -56,7 +56,8 @@ def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
     return stepSize, beta, Xbeta, func
 
 
-def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic):
+@normalize()
+def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     '''Michael Grant's implementation of Gradient Descent.'''
 
     loglike, gradient = family.loglike, family.gradient
@@ -112,7 +113,8 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic):
     return beta
 
 
-def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
+@normalize()
+def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     '''Newtons Method for Logistic Regression.'''
 
     gradient, hessian = family.gradient, family.hessian
@@ -150,6 +152,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
     return beta
 
 
+@normalize()
 def admm(X, y, regularizer=L1, lamduh=0.1, rho=1, over_relax=1,
          max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
@@ -243,7 +246,8 @@ def shrinkage(x, kappa):
     return z
 
 
-def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
+@normalize()
+def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic, **kwargs):
     '''Simple implementation of BFGS.'''
 
     n, p = X.shape
@@ -319,6 +323,7 @@ def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
     return beta
 
 
+@normalize()
 def proximal_grad(X, y, regularizer=L1, lamduh=0.1, family=Logistic,
                   max_iter=100, tol=1e-8, verbose=False):
 
@@ -383,7 +388,11 @@ def proximal_grad(X, y, regularizer=L1, lamduh=0.1, family=Logistic,
         stepSize *= stepGrowth
         backtrackMult = nextBacktrackMult
 
-    return beta
+    # L2-regularization returned a dask-array
+    try:
+        return beta.compute()
+    except AttributeError:
+        return beta
 
 
 _solvers = {
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -113,7 +113,6 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
 
 @normalize()
 def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
-    '''Newtons Method for Logistic Regression.'''
 
     gradient, hessian = family.gradient, family.hessian
     n, p = X.shape
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -56,7 +56,7 @@ def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
     return stepSize, beta, Xbeta, func
 
 
-@normalize()
+@normalize
 def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     '''Michael Grant's implementation of Gradient Descent.'''
 
@@ -111,7 +111,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     return beta
 
 
-@normalize()
+@normalize
 def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
 
     gradient, hessian = family.gradient, family.hessian
@@ -149,7 +149,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     return beta
 
 
-@normalize()
+@normalize
 def admm(X, y, regularizer='l1', lamduh=0.1, rho=1, over_relax=1,
          max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
@@ -237,7 +237,7 @@ def local_update(X, y, beta, z, u, rho, f, fprime, solver=fmin_l_bfgs_b):
     return beta
 
 
-@normalize()
+@normalize
 def lbfgs(X, y, regularizer=None, lamduh=1.0, max_iter=100, tol=1e-4,
           family=Logistic, verbose=False):
     """L-BFGS solver using scipy.optimize implementation"""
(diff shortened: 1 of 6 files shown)


TomAugspurger:
I'm going to ask some silly questions now, so apologies up front :) My optimization course didn't cover silly things like overflow/underflow.

Is normalization required to solve things like overflow? API-wise, I'd prefer that the user specifies this when instantiating their estimator, either through a new keyword or by using a pipeline. Then we don't have to worry about inferring whether the data has an intercept or not, since the user specified that as well. And the algorithms don't have to worry about it at all.

For solving the overflow issues, do we need something like https://lingpipe-blog.com/2012/02/16/howprevent-overflow-underflow-logistic-regression/? Or is your transformation and log1p enough?

moody-marlin:
No those are good questions; I'm fairly certain the `log1p` trick I did is the same thing that blog post mentions.

In general, normalization is a numerical stability trick, so in my mind it's a part of the algorithm stability and not really an API thing; I don't think a pipeline is necessary either because if you want to score a new input, you don't need to normalize it (because we remove the effect of normalization on the coefficient estimates at the end).  There's also a normalization I can apply in the absence of an intercept that I will add later this evening, and also update you on whether (or not) that fixes the issue you found.

^^^ everything I said above is opinion and open for discussion.

dask_glm/algorithms.py
```diff
@@ -97,22 +98,21 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic):
...
         stepSize *= stepGrowth
         backtrackMult = nextBacktrackMult
 
     return beta
 
 
-def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
+@normalize()
+def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     '''Newtons Method for Logistic Regression.'''
```
mpancia: Might want to change this comment -- this solver isn't specific to LR?

moody-marlin: Yea, very true; in general, we need to make a PR with actual doc strings...

moody-marlin:
@TomAugspurger I think I fixed the issue you found:

Off this PR:
```
import warnings

import dask.array as da
import numpy as np
from scipy.optimize import fmin_l_bfgs_b
from sklearn import datasets
from sklearn.linear_model import LogisticRegression
from dask_glm.algorithms import admm
from dask_glm.regularizers import L2

warnings.simplefilter("ignore", RuntimeWarning)

C = 10
lambda_ = 1 / C
np.random.seed(2)
X, y = datasets.make_classification(n_classes=2, n_samples=1000)

Xda = da.from_array(X, (10, X.shape[1]))
yda = da.from_array(y, (10,))

lrbase = LogisticRegression(fit_intercept=False, C=C, max_iter=10, penalty='l2').fit(X, y)
admm_base = admm(Xda, yda, lamduh=lambda_, max_iter=10, regularizer=L2)


def fit_and_compare(X, y, n):
    X = X.copy()
    np.random.seed(2)
    idx = np.random.choice(np.arange(len(X)), size=(n,))
    X[idx, 1] = np.random.randint(750, 1000, size=(n,))

    lr = LogisticRegression(fit_intercept=False, C=C, max_iter=10, penalty='l2')
    lr.fit(X, y)

    Xda = da.from_array(X, (10, X.shape[1]))
    yda = da.from_array(y, (10,))
    result = admm(Xda, yda, lamduh=lambda_, max_iter=10, regularizer=L2)

    print("Overflow={}".format(n)) # I don't use fancy new string formatting yet...
    print("Sklearn:", np.mean(np.abs(lrbase.coef_ - lr.coef_)))
    print("admm   :", np.mean(np.abs(admm_base - result)))
    return lr.coef_, result


fit_and_compare(X, y, 0)
fit_and_compare(X, y, 5)
fit_and_compare(X, y, 10)
fit_and_compare(X, y, 50)
fit_and_compare(X, y, 100)
``` 
gives output:
```
Overflow=0
Sklearn: 7.3964272207e-17
admm   : 1.66475297089e-13
Overflow=5
Sklearn: 0.00377521192273
admm   : 0.00649769477161
Overflow=10
Sklearn: 0.00311021598151
admm   : 0.00552952257074
Overflow=50
Sklearn: 0.00055956264663
admm   : 0.00479583189913
Overflow=100
Sklearn: 0.00312859460069
admm   : 0.00708823071006
```

and actually, the results are much worse for `sklearn` when using `l1` regularization.

moody-marlin:
@mrocklin any thoughts on why the [distributed scheduler determinism test](https://github.com/moody-marlin/dask-glm/blob/3f58be558df457b2be71ff06d85e35550bcd44a6/dask_glm/tests/test_algos_families.py#L157) fails [here](https://github.com/moody-marlin/dask-glm/blob/3f58be558df457b2be71ff06d85e35550bcd44a6/dask_glm/utils.py#L19) with a [`ValueError`](https://travis-ci.org/dask/dask-glm/jobs/227808777#L745-L751)?  It's trying to set some elements of a `numpy` array to 0, I'm not sure why the scheduler changes that.

dask_glm/algorithms.py
TomAugspurger: I started working on docs last night, I'll have something by the end of the week / early next week.

mrocklin:
Hrm, interesting.  We're getting this error `ValueError: assignment destination is read-only`, which is a sign that this particular numpy array is read-only.  I don't know why this would be offhand.  My lazy recommendation would be to make a copy of it if this occurs or to set it to writeable with http://stackoverflow.com/a/39554807/616616

dask_glm/utils.py
```diff
@@ -5,9 +5,32 @@
...
 
 
+def normalize(normalize=True):
+    def decorator(algo):
+        @wraps(algo)
+        def normalize_inputs(X, y, *args, **kwargs):
+            if normalize:
+                mean, std = da.compute(X.mean(axis=0), X.std(axis=0))
+                mean, std = mean.copy(), std.copy()
+                intercept_idx = np.where(std == 0)
```
TomAugspurger: If `X` has multiple columns with no standard deviation, does everything still work ok? You probably have bigger problems anyway, but does this cause any issues in the new code?

dask_glm/utils.py
```diff
@@ -5,9 +5,32 @@
...
 from multipledispatch import dispatch
 
 
+def normalize(normalize=True):
+    def decorator(algo):
+        @wraps(algo)
+        def normalize_inputs(X, y, *args, **kwargs):
+            if normalize:
+                mean, std = da.compute(X.mean(axis=0), X.std(axis=0))
+                mean, std = mean.copy(), std.copy()
```
TomAugspurger: Maybe leave a comment here about why the copies were necessary.

dask_glm/utils.py
moody-marlin: It will cause issues and return an incorrect result; I will add a check for that.

moody-marlin:
>  I think it if a user already knows that their data has been normalized (say in a pipeline), with keywords they could pass normalize=False and avoid a few computations.

This is a good point and convinces me the decorator solution isn't sufficient; I'll work on this and tag you whenever it's done.

moody-marlin:
@TomAugspurger `normalize` is now a `kwarg` for all algorithms.  I still left it as a decorator because normalization requires pre-algo and post-algo processing, and this seemed like the best way to handle it without reproducing code.  You can toggle this in your example and see the results with/without normalization:
```
import warnings

import dask.array as da
import numpy as np
from scipy.optimize import fmin_l_bfgs_b
from sklearn import datasets
from sklearn.linear_model import LogisticRegression
from dask_glm.algorithms import admm
from dask_glm.regularizers import L2

warnings.simplefilter("ignore", RuntimeWarning)

normalize=True
C = 10
lambda_ = 1 / C
np.random.seed(2)
X, y = datasets.make_classification(n_classes=2, n_samples=1000)

Xda = da.from_array(X, (10, X.shape[1]))
yda = da.from_array(y, (10,))

lrbase = LogisticRegression(fit_intercept=False, C=C, max_iter=10, penalty='l2').fit(X, y)
admm_base = admm(Xda, yda, lamduh=lambda_, max_iter=10, regularizer='l2', normalize=normalize)


def fit_and_compare(X, y, n):
    X = X.copy()
    np.random.seed(2)
    idx = np.random.choice(np.arange(len(X)), size=(n,))
    X[idx, 1] = np.random.randint(750, 1000, size=(n,))

    lr = LogisticRegression(fit_intercept=False, C=C, max_iter=10, penalty='l2')
    lr.fit(X, y)

    Xda = da.from_array(X, (10, X.shape[1]))
    yda = da.from_array(y, (10,))
    result = admm(Xda, yda, lamduh=lambda_, max_iter=10, regularizer='l2', normalize=normalize)

    print("Overflow={}".format(n)) # I don't use fancy new string formatting yet...
    print("Sklearn:", np.mean(np.abs(lrbase.coef_ - lr.coef_)))
    print("admm   :", np.mean(np.abs(admm_base - result)))
    return lr.coef_, result


fit_and_compare(X, y, 0)
fit_and_compare(X, y, 5)
fit_and_compare(X, y, 10)
fit_and_compare(X, y, 50)
fit_and_compare(X, y, 100)
```

dask_glm/utils.py
```diff
@@ -5,9 +5,35 @@
...
+                mean = mean if len(intercept_idx[0]) else np.zeros(mean.shape)
+                Xn = (X - mean) / std
+                out = algo(Xn, y, *args, **kwargs)
+                i_adj = np.sum(out * mean / std)
+                out[intercept_idx] -= i_adj
+                return out / std
+            else:
+                return algo(X, y, *args, **kwargs)
+        return normalize_inputs
+    return decorator
```
mrocklin: As a general comment I've found that decorators on public functions do come at some development cost.  It makes it somewhat harder to diagnose issues when debugging because the control flow is a bit harder to follow.

It may still be the best thing here.  I can't say.

dask_glm/utils.py
```diff
@@ -5,9 +5,35 @@
 
 import dask.array as da
 import numpy as np
+from functools import wraps
 from multipledispatch import dispatch
 
 
+def normalize():
```
TomAugspurger: I think since we don't use `@normalize` with any arguments, you can remove one level of nestedness here by removing the `def decorator(algo)` function. So like

```python
def normalize(algo):
    @wraps(algo)
    def normalize_inputs(...)
        ...
    return normalize_inputs
```

And then using it like `@normalize` instead of `@normalize()`

moody-marlin: Good call, done.

dask_glm/algorithms.py
```diff
@@ -272,8 +267,9 @@ def compute_loss_grad(beta, X, y):
     return beta
 
 
+@normalize
 def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
-                  max_iter=100, tol=1e-8, verbose=False):
+                  max_iter=100, tol=1e-8):
```
TomAugspurger: @moody-marlin just noticed, should proximal_grad accept `**kwargs` as well?

moody-marlin: Yea, actually all the algorithms should; it won't break anything now but should definitely be updated in whatever PR gets merged next.

TomAugspurger:
I can do it in my doc PR.

dask_glm/algorithms.py
TomAugspurger: 91cbda4 has it

Issue #9, opened 2017-01-26
Robustness Tests


cicdw:
We should have a testing framework / experimentation engine to test the algorithms for robustness to various types of input data (normalized / not-normalized / bad condition numbers / partial separability / low class sizes, etc.) -- eventually we might even want some sort of switching method that, for example, might detect bad conditioning in Newton's method and switch to gradient descent. This will require first writing a way to generate such data.

@mcg1969 @jcrist

mcg1969:
Once I get a bunch of delegatable tasks built out for my larger project I'm going to take a stab at this. Hoping to do so Wednesday or Thursday.

Issue #7, opened 2017-01-26
Logistic Accuracy tests


cicdw:
We need to determine a robust testing framework for our logistic regression algorithms, starting with a review of the accuracy of the current ADMM code.  

**Unregularized Problems**
- For unregularized problems with an intercept, we can start with the high-level test
```y.sum() = sigmoid(X.dot(beta)).sum()```
- For unregularized problems without an intercept, we can create a handful of tiny datasets to test on, or use well-studied public datasets (e.g., iris)

**Regularized Problems**
- As @hussainsultan already did, we can start with some high level 'marginal' tests which test extreme values for the various input parameters (e.g., when the regularization parameter is 'large' for l1-problems, the coefficients should all be 0)
- For regularized problems, we might need to hand craft a few tiny datasets to test on for accuracy

moody-marlin:
See #12

moody-marlin:
For unregularized problems, we can test for optimality of arbitrary problems easily by using `y = sigmoid(X.dot(beta))` (no thresholding) and testing that the estimated coefficients are close to `beta`;  how do we reliably test regularized problems though? I've considered dual certificates, and possibly 0 (sub)gradients, but is there a better way?

cc: @mcg1969

moody-marlin:
Some examples of the issues we face in implementing a robust testing framework can be found here: https://github.com/dask/dask-glm/blob/master/notebooks/AccuracyBook.ipynb

mcg1969:
Yeah, some sort of dual or subgradient criterion is likely needed...

moody-marlin:
https://gist.github.com/moody-marlin/e2de54ca17d615b263f80372031cb865 cc: @mpancia

Proximal grad does worst because the line search is currently very crude.

stoneyv:
I need to find support for this, but in Stephen P. Boyd's (Stanford EE Dept) presentation about ADMM on Jan 25th 2012 mentioned that ADMM without regularization is fragile.  What makes it very robust and guaranteed to converge is regularization.  This is mentioned somewhere before the 18 minute mark.  I will find a source for this.  http://videolectures.net/nipsworkshops2011_boyd_multipliers/

PR #46, opened 2017-05-02
Add Poisson regression support (WIP).

mpancia:
This adds an implementation of Poisson regression for count data, mimicking what is implemented for Logistic regression. I'm having some trouble debugging it, so I thought I'd submit this PR for a little help.

After testing it, the solver methods all give me wacky output except `newton`, which seems sane. The other methods give me division by zero errors in a logarithm, which I can't manage to track down. This seems to imply that there's something messed up in my implementation of the likelihood or something.

### DIFF
--- a/dask_glm/estimators.py
+++ b/dask_glm/estimators.py
@@ -5,9 +5,7 @@ from sklearn.base import BaseEstimator
 
 from . import algorithms
 from . import families
-from .utils import (
-    sigmoid, dot, add_intercept, mean_squared_error, accuracy_score
-)
+from .utils import (sigmoid, dot, add_intercept, mean_squared_error, accuracy_score, exp, poisson_deviance)
 
 
 class _GLM(BaseEstimator):
@@ -132,3 +130,51 @@ class LinearRegression(_GLM):
 
     def score(self, X, y):
         return mean_squared_error(y, self.predict(X))
+
+class PoissonRegression(_GLM):
+    """
+    Parameters
+    ----------
+    fit_intercept : bool, default True
+        Specifies if a constant (a.k.a. bias or intercept) should be
+        added to the decision function.
+    solver : {'admm', 'gradient_descent', 'newton', 'bfgs', 'proximal_grad'}
+        Solver to use. See :ref:`algorithms` for details
+    regularizer : {'l1', 'l2'}
+        Regularizer to use. See :ref:`regularizers` for details.
+        Only used with ``admm`` and ``proximal_grad`` solvers.
+    max_iter : int, default 100
+        Maximum number of iterations taken for the solvers to converge
+    tol : float, default 1e-4
+        Tolerance for stopping criteria. Ignored for ``admm`` solver
+    lambduh : float, default 1.0
+        Only used with ``admm`` and ``proximal_grad`` solvers
+    rho, over_relax, abstol, reltol : float
+        Only used with the ``admm`` solver. See :ref:`algorithms.admm`
+        for details
+
+    Attributes
+    ----------
+    coef_ : array, shape (n_classes, n_features)
+    intercept_ : float
+
+    Examples
+    --------
+    >>> from dask_glm.datasets import make_poisson
+    >>> X, y = make_poisson()
+    >>> pr = PoissonRegression()
+    >>> pr.fit(X, y)
+    >>> pr.predict(X)
+    >>> pr.get_deviance(X, y)
+    """
+
+    @property
+    def family(self):
+        return families.Poisson
+
+    def predict(self, X):
+        X_ = self._maybe_add_intercept(X)
+        return exp(dot(X_, self._coef))
+
+    def get_deviance(self, X, y):
+        return poisson_deviance(y, self.predict(X))
--- a/dask_glm/estimators.py
+++ b/dask_glm/estimators.py
@@ -5,7 +5,9 @@ from sklearn.base import BaseEstimator
 
 from . import algorithms
 from . import families
-from .utils import (sigmoid, dot, add_intercept, mean_squared_error, accuracy_score, exp, poisson_deviance)
+from .utils import (
+    sigmoid, dot, add_intercept, mean_squared_error, accuracy_score, exp, poisson_deviance
+)
 
 
 class _GLM(BaseEstimator):
@@ -120,6 +122,7 @@ class LinearRegression(_GLM):
     """
     Ordinary Lest Square regression
     """
+
     @property
     def family(self):
         return families.Normal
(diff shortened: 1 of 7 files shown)


dask_glm/estimators.py
```diff
@@ -5,9 +5,7 @@
 
 from . import algorithms
 from . import families
-from .utils import (
-    sigmoid, dot, add_intercept, mean_squared_error, accuracy_score
-)
+from .utils import (sigmoid, dot, add_intercept, mean_squared_error, accuracy_score, exp, poisson_deviance)
```
mpancia: Blame pycharm for cleaning this up.

dask_glm/families.py
```diff
@@ -38,7 +39,7 @@ def hessian(Xbeta, X):
 class Normal(object):
     @staticmethod
     def loglike(Xbeta, y):
-        return ((y - Xbeta)**2).sum()
+        return ((y - Xbeta) ** 2).sum()
```
mpancia: Similar pycharm PEPping.

dask_glm/utils.py
```diff
@@ -8,6 +8,7 @@
 from multipledispatch import dispatch
 
 
+@dispatch(np.ndarray)
```
mrocklin: I don't think that this is necessary.  I also think that this is causing some of your errors.

This function should work on all objects the same.  The different array classes are handled by dispatching on `exp` within this function.

dask_glm/families.py
mrocklin: From my perspective most of these changes are welcome.

dask_glm/families.py
```diff
@@ -59,3 +60,40 @@ def gradient(Xbeta, X, y):
...
+    @staticmethod
+    def pointwise_gradient(beta, X, y):
+        beta, y = beta.ravel(), y.ravel()
+        Xbeta = X.dot(beta)
+        return Poisson.gradient(Xbeta, X, y)
+
+    @staticmethod
+    def gradient(Xbeta, X, y):
+        eXbeta = exp(Xbeta)
+        return dot(X.T, eXbeta - y)
```
moody-marlin: This looks like the negative gradient to me.

moody-marlin: Actually, see comment above.

dask_glm/families.py
```diff
@@ -59,3 +60,40 @@ def gradient(Xbeta, X, y):
...
+    """
+    This implements Poisson regression for count data. 
+    See https://en.wikipedia.org/wiki/Poisson_regression.
+    """
+
+    @staticmethod
+    def loglike(Xbeta, y):
+        eXbeta = exp(Xbeta)
+        yXbeta = y*Xbeta
+        return (yXbeta - eXbeta).sum()
```
moody-marlin: Ah, actually this is probably the issue; what you have written appears to be the *true* log-likelihood (which makes sense considering we called this `loglike`).  In fact, this should be the *negative* log-likelihood because all of our functions attempt to minimize.

This is a failure on my part of documentation / naming.

moody-marlin:
So the problem is most likely due to sign error; the `loglike` should be the *negative* log-likelihood.  Your implementation of the gradient and the hessian *are* already for the negative log-likelihood, which explains why `newton` worked (it never calls the `loglike` method) and why all other algos failed. 

We might want to consider changing this method name altogether to `neg_loglike` or something...

cc: @TomAugspurger (for if you're working on doc strings).

mpancia:
@moody-marlin Well, changing the sign seems to make the output sane. I should have considered a sign error and looked at the other implementations, but I think that (like you mentioned) it might be a good idea to rename the method to `neg_loglike` or at least having a comment that gives the appropriate caveat. :)

stoneyv:
It would be good to include a quasi-poisson regression method (quasi likelihood Wedderburn 1974) similar to the R glm package as well as generalized estimate equations like the R "sandwich" package.  You need these functions to check how close the mean is to the variance to determine the robustness of your poisson regression model as an estimator.  Is this non linear estimator better than adding a squared term to the linear model or using a negative binomial model when the variance is less or greater than the mean.  Poisson regression may be a good choice for unbounded counts, rates or proportions.  
Blog article about using quasi-poisson for count data with variance greater than the mean  
http://www.theanalysisfactor.com/glm-r-overdispersion-count-regression/  
Sandwich agnostic variance (generalized estimate equations)  
https://www.jstatsoft.org/article/view/v016i09  
https://cran.r-project.org/web/packages/sandwich/sandwich.pdf

moody-marlin:
@stoneyv I agree that over-dispersion is an issue and quasi-likelihood methods are worthy of consideration -- maybe you could raise an issue for either including quasi-likelihood based families or a `metrics` module that helps diagnose when quasi-likelihood methods are needed?  

I don't think over-dispersion metrics / implementations should be included in this PR though - IMO `PoissonRegression` should avoid interpreting the user's intent and implement straightforward Poisson Regression, which I think is what this PR does.

moody-marlin:
@mpancia if you add the `Poisson` family to the parametrized tests in `test_algos_families.py`, and then flake the code I think we'll be good to go here.

dask_glm/families.py
```diff
@@ -59,3 +60,40 @@ def gradient(Xbeta, X, y):
...
+    @staticmethod
+    def gradient(Xbeta, X, y):
+        eXbeta = exp(Xbeta)
+        return dot(X.T, eXbeta - y)
+
+    @staticmethod
+    def hessian(Xbeta, X):
+        eXbeta = exp(Xbeta)
+        diag_eXbeta = da.diag(eXbeta)
+        x_diag_eXbeta = dot(diag_eXbeta, X)
```
moody-marlin: FWIW I think `eXbeta[:, None] * X` should broadcast appropriately and save some memory (avoids storing all the `0`'s from the `diag` creation).

mpancia: True. Weird `numpy` idioms abound.

mpancia:
@moody-marlin Looks good, though it seems a bit weird as @mrocklin mentioned to be using the logistic test data from `make_y` to test the normal LR or the Poisson regression - some family-specific tests based on appropriately-generated data might make sense.

At the same time, those are tests for gradient descent working, so regardless of whether or not the family is appropriate for the actual data, the descent should decrease loss. *shrug*

moody-marlin:
@mpancia Ah yea good call out; ultimately this re-raises the question of how to make a truly robust GLM testing framework (#7, #9, etc.).  However, I think that opens a new can of worms that is beyond the scope of this PR.  I think this provides an informative (and useful) example for how to include future GLM families into `dask-glm`.

LGTM; I'll let this hang out until tomorrow in case anyone else has further feedback.

mpancia:
@moody-marlin Great, thanks. The Travis build is passing now, BTW; might want to squash this when you merge so as not to muss up the main commit history, or I can rebase it to get a cleaner commit history.

PR #49, opened 2017-05-03
Elastic Net Regularizer

### LINKED ISSUE #47: L2 regularization scaling is inconsistent
If we use define `l2` regularization as:
```
min_x ( f(x) + lambda / 2 \|x\|_2^2 )
```
then the regularizer is given by
```
1/2 \|x\|_2^2
```
which has proximal operator `1 / (1 +t) x` (which is what we [currently have implemented](https://github.com/dask/dask-glm/blob/master/dask_glm/regularizers.py#L10)).  However, all of our [other methods](https://github.com/dask/dask-glm/blob/master/dask_glm/regularizers.py#L24) assume the regularizer is given by `\|x\|_2^2` (without the 1/2 scaling).  

This doesn't affect quality, but we should be consistent.

cc @postelrich

### LINKED ISSUE #48: Abstract Base Classes
We should probably implement abstract base classes for `Regularizer` and `Family` with corresponding documentation; as #47 and #46 have highlighted, the lack of a solid template can lead to problems.

- [x] Regularizer Base Class
- [ ] Family Base Class

postelrich:
This PR adds elastic net regularization, the weighted sum of l1 and l2, as an option.  Couldn't find a proximal operator so derived it with much much help from @moody-marlin in an included notebook. It also adds an abstract base class for `Regularizer` and fixes the formulas for l2.

Resolves #48 
Resolves #47

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -150,7 +150,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
     return beta
 
 
-def admm(X, y, regularizer=L1, lamduh=0.1, rho=1, over_relax=1,
+def admm(X, y, regularizer=L1(), lamduh=0.1, rho=1, over_relax=1,
          max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
@@ -319,7 +319,7 @@ def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
     return beta
 
 
-def proximal_grad(X, y, regularizer=L1, lamduh=0.1, family=Logistic,
+def proximal_grad(X, y, regularizer=L1(), lamduh=0.1, family=Logistic,
                   max_iter=100, tol=1e-8, verbose=False):
 
     n, p = X.shape
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -24,7 +24,7 @@ from scipy.optimize import fmin_l_bfgs_b
 
 from dask_glm.utils import dot, exp, log1p
 from dask_glm.families import Logistic
-from dask_glm.regularizers import L1, _regularizers
+from dask_glm.regularizers import _regularizers
 
 
 def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
@@ -150,7 +150,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
     return beta
 
 
-def admm(X, y, regularizer=L1(), lamduh=0.1, rho=1, over_relax=1,
+def admm(X, y, regularizer='l1', lamduh=0.1, rho=1, over_relax=1,
          max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
@@ -319,7 +319,7 @@ def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
     return beta
 
 
-def proximal_grad(X, y, regularizer=L1(), lamduh=0.1, family=Logistic,
+def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
                   max_iter=100, tol=1e-8, verbose=False):
 
     n, p = X.shape
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -24,7 +24,7 @@ from scipy.optimize import fmin_l_bfgs_b
 
 from dask_glm.utils import dot, exp, log1p
 from dask_glm.families import Logistic
-from dask_glm.regularizers import _regularizers
+from dask_glm.regularizers import Regularizer
 
 
 def compute_stepsize_dask(beta, step, Xbeta, Xstep, y, curr_val,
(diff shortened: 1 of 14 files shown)


dask_glm/algorithms.py
```diff
@@ -319,7 +319,7 @@ def bfgs(X, y, max_iter=500, tol=1e-14, family=Logistic):
     return beta
 
 
-def proximal_grad(X, y, regularizer=L1, lamduh=0.1, family=Logistic,
+def proximal_grad(X, y, regularizer=L1(), lamduh=0.1, family=Logistic,
```
TomAugspurger: Thoughts about changing this to `'l1`' and looking it up in `_regularizers`?

postelrich: I agree, should be change.  Also in addition, we could move the string logic to a function in `regularizers.py`

postelrich:
@TomAugspurger What do you think of my latest commit?  I use the base to retrieve child classes via string.

dask_glm/regularizers.py
```diff
@@ -9,6 +9,7 @@ class Regularizer(object):
     Defines the set of methods required to create a new regularization object. This includes
     the regularization functions itself and it's gradient, hessian, and proximal operator.
     """
+    _name = '_base'
```
TomAugspurger: I think maybe this should be just `name`. I think @moody-marlin mentioned that users might defining their own regularizers, so they shouldn't need to override private attributes.

dask_glm/algorithms.py
```diff
@@ -150,12 +150,12 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic):
...
 
 
-def admm(X, y, regularizer=L1, lamduh=0.1, rho=1, over_relax=1,
+def admm(X, y, regularizer='l1', lamduh=0.1, rho=1, over_relax=1,
          max_iter=250, abstol=1e-4, reltol=1e-2, family=Logistic):
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
-    regularizer = _regularizers.get(regularizer, regularizer)  # string
+    regularizer = Regularizer.get(regularizer)
```
moody-marlin: I'll be sad to see this line go, but :+1:

dask_glm/regularizers.py
```diff
@@ -3,85 +3,123 @@
...
 
-class L2(object):
+class Regularizer(object):
+    """Abstract base class for regularization object.
 
-    @staticmethod
-    def proximal_operator(beta, t):
-        return 1 / (1 + t) * beta
+    Defines the set of methods required to create a new regularization object. This includes
+    the regularization functions itself and it's gradient, hessian, and proximal operator.
```
moody-marlin: it's -> its

dask_glm/regularizers.py
```diff
@@ -3,85 +3,123 @@
...
+
+    def hessian(self, beta):
+        """Hessian of regularization function."""
+        raise NotImplementedError
 
-    @staticmethod
-    def f(beta):
-        return (beta**2).sum()
+    def proximal_operator(self, beta, t):
+        """Proximal operator function for non-differentiable regularization function."""
```
moody-marlin: ```Proximal operator for regularization function```; the regularizer doesn't need to be non-differentiable, it just *can* be.

dask_glm/regularizers.py
```diff
@@ -3,85 +3,123 @@
...
         else:
             return np.sign(beta)
 
-    @staticmethod
-    def add_reg_grad(grad, lam):
-        def wrapped(beta, *args):
-            return grad(beta, *args) + lam * L1.gradient(beta)
-        return wrapped
+    def hessian(self, beta):
+        raise ValueError('l1 norm is not twice differentiable!')
```
moody-marlin: We should probably just fix this with this PR; this should be similar to the gradient:
```
if np.any(np.isclose(beta, 0)):
    raise ValueError('l1 norm is not twice differentiable at 0!')
else:
    return np.zeros((beta.shape[0], beta.shape[0]))
```
l1 regularization is a straight line everywhere except at 0 where there's a kink.

dask_glm/tests/test_algos_families.py
```diff
@@ -89,7 +89,7 @@ def test_basic_unreg_descent(func, kwargs, N, nchunks, family):
 @pytest.mark.parametrize('nchunks', [1, 10])
 @pytest.mark.parametrize('family', [Logistic, Normal, Poisson])
 @pytest.mark.parametrize('lam', [0.01, 1.2, 4.05])
-@pytest.mark.parametrize('reg', [L1, L2])
+@pytest.mark.parametrize('reg', [r() for r in Regularizer.__subclasses__()])
```
moody-marlin: Nice.

dask_glm/regularizers.py
postelrich: I agree, made the change.

dask_glm/regularizers.py
postelrich: Then do we want to switch the elastic net hessian to include the l1 side to the weight?  It won't have an effect except raise when there are errors.

moody-marlin: Ah yes, definitely - I missed that.

postelrich:
@moody-marlin ok implemented your changes, looks good to go.

moody-marlin:
@TomAugspurger do you have any other comments / suggestions?  LGTM.

TomAugspurger:
👍  on a quick skim.

PR #50, opened 2017-05-05
L-BFGS solver based on scipy.optimize with L2 regularization

MLnick:
Refer to #30 for the discussion and prototype from which this was developed.

Still a little bit WIP, though it supports `L2` regularizer and:
* is equivalent to sklearn's `LogisticRegression` with `solver=lbfgs` and `fit_intercept=False`
* doesn't give precisely the same solution as sklearn's `Ridge` for squared loss (as the solvers are different), but achieves essentially the same RMSE metric on some simple test data (`fit_intercept=False`)

Todos:
- [ ] test cases
- [ ] maybe some performance testing (refer e.g. https://gist.github.com/mrocklin/4e486064882cce630ffb4ee4e39bc333)
- [ ] support `L1` regularizer (via adapting https://gist.github.com/vene/fab06038a00309569cdd). Could this potentially also support `ElasticNet` regularizer?

**Note** not sure if the estimator versions will give the same results as sklearn because intercept handling may be different. Still need to investigate.

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -243,14 +243,15 @@ def shrinkage(x, kappa):
     return z
 
 
-def lbfgs(X, y, max_iter=100, tol=1e-4, alpha=1.0, family=Logistic, verbose=False):
+def lbfgs(X, y, regularizer='l1', alpha=1.0, max_iter=100, tol=1e-4,
+          family=Logistic, verbose=False):
     """L-BFGS solver using scipy.optimize implementation
     
     Currently supports L2 regularization only"""
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
-    regularizer = Regularizer.get('l2')
+    regularizer = Regularizer.get(regularizer)
 
     n, p = X.shape
     beta0 = np.zeros(p)
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -243,29 +243,30 @@ def shrinkage(x, kappa):
     return z
 
 
-def lbfgs(X, y, regularizer='l2', alpha=1.0, max_iter=100, tol=1e-4,
+def lbfgs(X, y, regularizer=None, alpha=1.0, max_iter=100, tol=1e-4,
           family=Logistic, verbose=False):
     """L-BFGS solver using scipy.optimize implementation"""
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
-    regularizer = Regularizer.get(regularizer)
+    if regularizer:
+        regularizer = Regularizer.get(regularizer)
+        pointwise_loss = regularizer.add_reg_f(pointwise_loss, alpha)
+        pointwise_gradient = regularizer.add_reg_grad(pointwise_gradient, alpha)
 
     n, p = X.shape
     beta0 = np.zeros(p)
 
-    def compute_loss_grad(beta, X, y, alpha):
-        loss_fn = regularizer.add_reg_f(pointwise_loss, alpha)(beta, X, y)
-        gradient_fn = regularizer.add_reg_grad(pointwise_gradient, alpha)(beta, X, y)
+    def compute_loss_grad(beta, X, y):
+        loss_fn = pointwise_loss(beta, X, y)
+        gradient_fn = pointwise_gradient(beta, X, y)
         loss, gradient = compute(loss_fn, gradient_fn)
         return loss, gradient.copy()
 
-    func = lambda x, *args: compute_loss_grad(x, *args)[0:2]
-
     with set_options(fuse_ave_width=0):  # optimizations slows this down
         beta, loss, info = fmin_l_bfgs_b(
-            func, beta0, fprime=None,
-            args=(X, y, alpha),
+            compute_loss_grad, beta0, fprime=None,
+            args=(X, y),
             iprint=(verbose > 0) - 1, pgtol=tol, maxiter=max_iter)
 
     return beta
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -243,11 +243,9 @@ def shrinkage(x, kappa):
     return z
 
 
-def lbfgs(X, y, regularizer='l1', alpha=1.0, max_iter=100, tol=1e-4,
+def lbfgs(X, y, regularizer='l2', alpha=1.0, max_iter=100, tol=1e-4,
           family=Logistic, verbose=False):
-    """L-BFGS solver using scipy.optimize implementation
-
-    Currently supports L2 regularization only"""
+    """L-BFGS solver using scipy.optimize implementation"""
 
     pointwise_loss = family.pointwise_loss
     pointwise_gradient = family.pointwise_gradient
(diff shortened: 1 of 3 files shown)


MLnick:
cc @mrocklin @moody-marlin @TomAugspurger

MLnick:
Script to produce the sklearn comparisons:

```python
import dask
import dask.array as da
from dask_glm.algorithms import lbfgs
from dask_glm.families import Logistic, Normal
import numpy as np
from sklearn import datasets
from sklearn.linear_model import LogisticRegression, Ridge
from sklearn.metrics import roc_auc_score, mean_squared_error

# compare with sklearn's LogisticRegression using L-BFGS solver
# make dataset
X, y = datasets.make_classification(n_classes=2, n_samples=10000, n_features=100)
# dask versions
Xda = da.from_array(X, 1000)
yda = da.from_array(y, 1000)

lr = LogisticRegression(fit_intercept=False, solver="lbfgs", max_iter=100)

betas_sklearn = []
betas_dask = []
for c in [1e5, 10.0, 1.0, 0.01]:
	lr.C = c
	lr.fit(X, y)
	betas_sklearn.append(lr.coef_)
	alpha = 1/c
	dask_beta = lbfgs(Xda, yda, alpha=alpha, max_iter=100, family=Logistic)
	betas_dask.append(dask_beta)

assert np.alltrue([np.allclose(a, b) for a, b in zip(betas_sklearn, betas_dask)]), "Coefficients for Logistic differ between sklearn and dask-glm"

# compare with sklearn's Ridge regressor
# make dataset
Xr, yr = datasets.make_regression(n_samples=10000, n_features=100)
# dask versions
Xrda = da.from_array(Xr, 1000)
yrda = da.from_array(yr, 1000)
ridge = Ridge(fit_intercept=False)

betas_sklearn_ridge = []
betas_dask_ridge= []
for l in [1e-10, 1e-3, 1e-1, 1.0, 10]:
	ridge.alpha = l
	ridge.fit(Xr, yr)
	betas_sklearn_ridge.append(ridge.coef_)
	dask_beta = lbfgs(Xrda, yrda, alpha=l, max_iter=100, family=Normal)
	betas_dask_ridge.append(dask_beta)

	ridge2 = Ridge(fit_intercept=False)
	ridge2.coef_ = dask_beta
	ridge2.intercept_ = 0.0
	print("alpha=%2.10f, RMSE Ridge=%2.6f, RMSE dask-glm=%2.6f" % (l, np.sqrt(mean_squared_error(y, ridge.predict(Xr))), np.sqrt(mean_squared_error(y, ridge2.predict(Xr)))))

```

dask_glm/algorithms.py
```diff
@@ -243,78 +243,31 @@ def shrinkage(x, kappa):
...
-
-        beta = beta - stepSize * step  # tiny bit of repeat work here to avoid communication
-        Xbeta = Xbeta - stepSize * Xstep
-
-        if stepSize == 0:
-            print('No more progress')
-            break
+    pointwise_loss = family.pointwise_loss
+    pointwise_gradient = family.pointwise_gradient
+    regularizer = Regularizer.get('l2')
```
moody-marlin: Could you elevate `regularizer` to a keyword?  On first glance it seems like this would work perfectly fine with `regularizer=None` or any other regularizer that is differentiable.  If the regularizer classes implement their methods appropriately, then if an algorithm runs into a point of non-differentiability, an error is thrown.

mrocklin:
I've taken the liberty of adding `lbfgs` to the testing suite in https://github.com/dask/dask-glm/pull/50/commits/8c1345b76af1d203d30aab33ee80365b732af5be .  Hope you don't mind me pushing to your fork @MLnick .  

It looks like most things work well, a couple things are failing tolerance checks.

mrocklin:
I ran this on a slightly larger problem with the distributed scheduler (just on my local laptop) and got [this profile](https://rawgit.com/mrocklin/d27442d02c0edfa5e1a049beada00a4c/raw/b063fdda6d3029fef4abbc064ba610556eef041c/task-stream-lbfgs.html).  I recommend clicking the wheel zoom tool and then zooming into an arbitrary part of the computation.

We're still mostly bound by whitespace / dead time.  We can resolve a lot of this just by using larger chunks, but this does motivate further work in optimizations.  I think that we could dramatically reduce overhead here if we did two things:

1.  Fused with a wider width (something like 10)
2.  Produced the graph only once, rather than at every iteration

You've correctly turned off fusion for the computation and I have no doubt that this will speed things up, especially on small chunks.  We should make it easier to produce the graph once though.  I think that this will allow us to be much more aggressive about optimization in the future.  This is something that I probably need to do within dask proper.

After this it looks like we're just bound by tensordot and exp, which is pretty typical.

Anyway, nothing for you to do here, I just thought you might enjoy seeing the profile plot and hearing about possibilities to make this faster in the future.

moody-marlin:
@mrocklin it looks like `lbfgs` is currently failing the tolerance checks just because `l2`-regularization is hard-coded into it and the tests are testing for unregularized problems; should be a quick fix.

MLnick:
@mrocklin thanks for pushing the test integration. For the fusion I just followed your example notebook referenced in #30 where you mentioned it would be better for performance.

@moody-marlin I can make `regularizer` a keyword arg, sure.

MLnick:
I still see some tolerance-related test failures locally in `dask_glm/tests/test_algos_families.py`/`test_methods`

moody-marlin:
@MLnick it looks like `test_methods` is failing -- this test is testing for accuracy on algorithms which it assumes are defaulting to solving unregularized problems.  If you change the default of `lbfgs` to `regularizer=None` (and update [these lines](https://github.com/MLnick/dask-glm/blob/dfdefdd1f6644077c6dc793b6f3c64900b90eeb1/dask_glm/algorithms.py#L258-L259) to not do anything) I expect the test to pass.

This raises a good issue though - we might need to update the tests to inspect for a `regularizer` kwarg and adapt accordingly.

mrocklin:
> This raises a good issue though - we might need to update the tests to inspect for a regularizer kwarg and adapt accordingly.

We could add in kwargs to the pytest parameters along with the algorithm like we do in some other tests.

moody-marlin:
Yea that could work well; the only added complexity is that there are some algorithm / regularizer pairs that won't work together (example: `newton` and `l1`) so we would need to handle that appropriately.

MLnick:
@moody-marlin ah right. I agree it makes most sense to set the regularizer
default to None and adapt the lbfgs code accordingly.

MLnick:
@mrocklin by the way, I keep getting the following warning for `lbfgs`: 

```
.../dask/array/core.py:476: RuntimeWarning: overflow encountered in true_divide
```

mrocklin:
Looks like tests are passing :)

MLnick:
Do we need any further tests?

moody-marlin:
I think we're good on tests for now but we should continue to think about how we can make them more comprehensive and leave that for a future PR.

LGTM!

mrocklin:
Merged.  Thanks @MLnick !  I'm looking forward to playing with this.

MLnick:
Awesome, thanks!

By the way any thoughts on changing "lamduh" -> "alpha" for consistency
with e.g. sklearn? :)

MLnick:
I will also try to do some larger-scale testing on my cluster :)

PR #55, opened 2017-05-17
ABC for Families

postelrich:
This PR adds an abstract base class for distribution families.  To provide the same way to track subclasses as Regularizers, I added a RegistryClass that both abc's can inherit.


dask_glm/families.py
```diff
@@ -1,18 +1,48 @@
...
     .. _Logistic regression: https://en.wikipedia.org/wiki/Logistic_regression
     """
-    @staticmethod
-    def loglike(Xbeta, y):
+    name = 'logistic'
+
+    def loglikelihood(self, Xbeta, y):
         """
-        Evaluate the logistic loglikeliehood
+        Evaluate the logistic loglikelihoodlihood
```
moody-marlin: spelling.  Also, this led to some confusion for the Poisson family so could you call out that this is the *negative* log-likelihood for all the families?

postelrich:
@moody-marlin @TomAugspurger feedback before I add tests?

moody-marlin:
@postelrich are you planning on adding more tests, or are you comfortable merging?

postelrich:
@moody-marlin im going to add tests

dask_glm/tests/test_algos_families.py
```diff
@@ -8,9 +8,33 @@
 
 from dask_glm.algorithms import (newton, lbfgs, proximal_grad,
                                  gradient_descent, admm)
-from dask_glm.families import Logistic, Normal, Poisson
+from dask_glm.families import Family
 from dask_glm.regularizers import Regularizer
 from dask_glm.utils import sigmoid, make_y
+from unittest.mock import patch
```
postelrich: @TomAugspurger yea the point was just to make sure that if someone changes this code, the tests will catch it since it would break every subclass would potentially break.

moody-marlin:
@postelrich want to revisit this and get the tests passing?

PR #56, opened 2017-05-22
Use setuptools scm

TomAugspurger:


postelrich:
@TomAugspurger I'm not familiar with setuptools_scm, how does it compare to versioneer?

TomAugspurger:
Seems to be "blessed" by the PyPA https://github.com/pypa/setuptools_scm, so that's good enough reason for me :D

postelrich:
Looks good to me.

moody-marlin:
LGTM.

PR #70, opened 2018-04-09
change to Numpy power ufunc

edenbaus:
Change builtin python power function to Numpy power ufunc

### DIFF
--- a/docs/examples/sigmoid.ipynb
+++ b/docs/examples/sigmoid.ipynb
@@ -47,7 +47,7 @@
     "    fpp = fp*(1-2*fo)\n",
     "    fppp = fpp*(1-4*fo)\n",
     "    d = x-xo\n",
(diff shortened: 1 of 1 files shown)


mrocklin:
Hi @edenbaus .  Thanks for the PR.  I'm a bit curious about this, I would have expected these to be more-or-less the same.  Is there a performance difference?

edenbaus:
Hello,

Thanks for looking at my pull request.

I looked into the performance differences. I'm a bit surprised, but the
base python power function tends to be faster than the numpy power function
for power = 2. However, when raising values to a power > 2, the performance
between the base python power function and the numpy power function is
effectively identical.

The other difference is precision, numpy's power function offers much
greater floating point precision over the base python power function.

Given the negligible performance difference between both approaches, it's
your call if you think the increased precision is important enough to
include this pull request.

Best Regards,
Scott Edenbaum

mrocklin:
My guess would be that `**` dispatches to `np.power` for numpy arrays.

PR #72, opened 2018-10-23
Update to use dask.config.set and schduler keyword

jrbourbeau:
This PR replaces instances of `dask.set_options` with `dask.config.set` and replaces usage of the `get` keyword with the `sceduler` keyword. 

Previously these would raise warnings, but https://github.com/dask/dask/pull/4077 elevated these warning to errors.

### DIFF
--- a/dask_glm/tests/test_estimators.py
+++ b/dask_glm/tests/test_estimators.py
@@ -64,7 +64,7 @@ def test_lm(fit_intercept):
 @pytest.mark.parametrize('fit_intercept', [True, False])
 def test_big(fit_intercept):
     import dask
-    dask.set_options(get=dask.get)
+    dask.config.set(scheduler='synchronous')
     X, y = make_classification()
     lr = LogisticRegression(fit_intercept=fit_intercept)
     lr.fit(X, y)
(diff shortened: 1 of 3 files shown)


dask_glm/tests/test_estimators.py
```diff
@@ -64,7 +64,7 @@ def test_lm(fit_intercept):
 @pytest.mark.parametrize('fit_intercept', [True, False])
 def test_big(fit_intercept):
     import dask
-    dask.set_options(get=dask.get)
+    dask.config.set(scheduler='synchronous')
```
TomAugspurger: Hmm, while we're at it could you update these to use a context manager?

TomAugspurger:
Thanks for doing this. I'll make a release after merging.

dask_glm/tests/test_estimators.py
jrbourbeau: For sure, good idea

jrbourbeau: Also, should I move `import dask` outside the tests? There's probably a reason for it being inside the test, but it's not clear to me

TomAugspurger:
Sure, fine with moving the import up.

PR #66, opened 2017-11-16
Deprecate estimators

TomAugspurger:
Closes #63 

xref https://github.com/dask/dask-ml/pull/94

PR #75, opened 2019-03-04
[WIP] Adjust all algorithms to work with CuPy

### LINKED ISSUE #73: NEP-18: CuPy backend requires unimplemented algorithms
For the past few weeks, I've been working on issues related to [NEP-18](https://www.numpy.org/neps/nep-0018-array-function-protocol.html) support for Dask (and ultimately, Dask-GLM as well) to allow CuPy to be used as a backend. I've made some progress, which I'll soon share with more details.

Among the issues I have now is the usage of the following Dask-GLM algorithms with CuPy:

1. `newton`: requires `[numpy.linalg.lstsq](https://www.numpy.org/devdocs/reference/generated/numpy.linalg.lstsq.html#numpy.linalg.lstsq)`
1. `admm` and `lbfgs`: require `[scipy.optimize.fmin_l_bfgs_b](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_l_bfgs_b.html)`

Both of these functions are not implemented in CuPy. So I'm wondering if anybody knows whether any of these functions have some open and reliable CUDA implementation that could be integrated into CuPy or if we could implement them somewhat easily on top of already supported CuPy functionality?

pentschev:
Requirements:
* https://github.com/numpy/numpy/pull/13046
* https://github.com/cupy/cupy/pull/2079

Resolves #73.

### DIFF
--- a/dask_glm/utils.py
+++ b/dask_glm/utils.py
@@ -5,6 +5,7 @@ import sys
 
 import dask.array as da
 import numpy as np
+import cupy
 from functools import wraps
 from multipledispatch import dispatch
 
@@ -21,7 +22,7 @@ def normalize(algo):
                 raise ValueError('Multiple constant columns detected!')
             mean[intercept_idx] = 0
             std[intercept_idx] = 1
-            mean = mean if len(intercept_idx[0]) else np.zeros(mean.shape)
+            mean = mean if len(intercept_idx[0]) else np.zeros_like(mean)
             Xn = (X - mean) / std
             out = algo(Xn, y, *args, **kwargs).copy()
             i_adj = np.sum(out * mean / std)
@@ -39,7 +40,7 @@ def sigmoid(x):
 
 @dispatch(object)
 def exp(A):
-    return A.exp()
+    return np.exp(A)
 
 
 @dispatch(float)
@@ -74,7 +75,7 @@ def absolute(A):
 
 @dispatch(object)
 def sign(A):
-    return A.sign()
+    return np.sign(A)
 
 
 @dispatch(np.ndarray)
@@ -89,7 +90,7 @@ def sign(A):
 
 @dispatch(object)
 def log1p(A):
-    return A.log1p()
+    return np.log1p(A)
 
 
 @dispatch(np.ndarray)
@@ -105,8 +106,7 @@ def log1p(A):
 @dispatch(object, object)
 def dot(A, B):
     x = max([A, B], key=lambda x: getattr(x, '__array_priority__', 0))
-    module = package_of(x)
-    return module.dot(A, B)
+    return np.dot(A, B)
 
 
 @dispatch(da.Array, np.ndarray)
@@ -139,6 +139,10 @@ def sum(A):
 @dispatch(np.ndarray)
 def add_intercept(X):
     return np.concatenate([X, np.ones((X.shape[0], 1))], axis=1)
+
+@dispatch(cupy.ndarray)
+def add_intercept(X):
+    return np.concatenate([X, np.ones_like(X, shape=(X.shape[0], 1))], axis=1)
 
 
 @dispatch(da.Array)
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -11,6 +11,7 @@ import dask.array as da
 from scipy.optimize import fmin_l_bfgs_b
 
 
+from dask.array.utils import normalize_to_array
 from dask_glm.utils import dot, normalize
 from dask_glm.families import Logistic
 from dask_glm.regularizers import Regularizer
@@ -97,7 +98,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros(p)
+    beta = np.zeros_like(X, shape=(p,))
 
     for k in range(max_iter):
         # how necessary is this recalculation?
@@ -161,7 +162,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     """
     gradient, hessian = family.gradient, family.hessian
     n, p = X.shape
-    beta = np.zeros(p)  # always init to zeros?
+    beta = np.zeros_like(X, shape=(p,))  # always init to zeros?
     Xbeta = dot(X, beta)
 
     iter_count = 0
@@ -178,8 +179,10 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
 
         # should this be dask or numpy?
         # currently uses Python 3 specific syntax
-        step, _, _, _ = np.linalg.lstsq(hess, grad)
-        beta = (beta_old - step)
+        step, _, _, _ = np.linalg.lstsq(normalize_to_array(hess), normalize_to_array(grad))
+        step_like = np.empty_like(X, shape=step.shape)
+        step_like[:] = step
+        beta = (beta_old - step_like)
 
         iter_count += 1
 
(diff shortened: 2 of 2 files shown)


pentschev:
The build will fail now, since it depends on NumPy and CuPy fixes that weren't yet merged. However, I can confirm that all existing tests pass on my setup after the changes here.

dask_glm/utils.py
```diff
@@ -164,6 +164,11 @@ def make_y(X, beta=np.array([1.5, -3]), chunks=2):
     y = da.random.random(z0.shape, chunks=z0.chunks) < sigmoid(z0)
     return y
 
+def cupy_make_y(X, beta=cupy.array([1.5, -3])):
+    n, p = X.shape
+    z0 = X.dot(beta)
+    y = cupy.random.random(z0.shape) < sigmoid(z0)
+    return y
```
mrocklin: We probably can't have this library `import cupy` at the top level, otherwise users will get an ImportError if they don't have cupy installed.

dask_glm/tests/cupy/test_regularizers.py
```diff
@@ -0,0 +1,130 @@
+import numpy as np
+import cupy
+import pytest
+import cupy.testing as cpt
+from dask_glm import regularizers as regs
```
mrocklin: I recommend

```python
import pytest
cupy = pytest.importorskip('cupy')
```

dask_glm/utils.py
mrocklin: We might move this into the `dask_glm/tests/cupy` module and import from there.

pentschev: > Also, I'm surprised that there weren't any code changes. All I'm seeing here is testing code (which all looks quite nice). Did you have to modify any code to get things to work?

There were a few code changes, in the first commit , mostly regarding `*_like()` calls and such. Did you see that?

https://github.com/dask/dask-glm/pull/75/commits/e39aafd289193df56cb7e254a4aa174d3dd2dcff

pentschev: Yes, I was planning to cleanup this cupy imports later, but I hadn't yet thought of how I should do it, your idea sounds good, I will do that.

mrocklin: Ah indeed, I must have accidentally been looking only at the most recent commit.

dask_glm/utils.py
```diff
@@ -140,6 +140,10 @@ def sum(A):
 def add_intercept(X):
     return np.concatenate([X, np.ones((X.shape[0], 1))], axis=1)
 
+@dispatch(cupy.ndarray)
```
pentschev: What about this cupy dispatch, do you think we should move it elsewhere?

mrocklin: Yes, the dask-glm library needs to remain usable even if users don't have cupy installed.  Most users won't have a GPU.  

Perhaps we can generalized the NumPy solution above to look more like this, and then replace the dispatch type of that function with `object` rather `np.ndarray` so that it covers both.  This will also get triggered then if this function is called on non-array objects, but that would be an error anyway.

mrocklin: This might be a common theme for other functions in this file as well.  We might try to find general implementations for the other functions as well.  

Looking at this code my *guess* is that the other `object` dispatches here are trying to support scipy.sparse matrices like 

```python
import scipy.sparse
x = scipy.sparse.csr_matrix([[1, 2], [0, 1]])
```

In some cases these extra dispatches are unnecessary, as in this case:

```python
@dispatch(object)
def log1p(A):
    return A.log1p()


@dispatch(np.ndarray)
def log1p(A):
    return np.log1p(A)

# where really the numpy implementation would have worked for either
In [11]: np.log1p(x)
Out[11]:
<2x2 sparse matrix of type '<class 'numpy.float64'>'
	with 3 stored elements in Compressed Sparse Row format>
```

In these cases we can probably replace the dispatched function with just plain `np.log1p`.  The `absolute` function can be replaced with just `abs` (Python dispatches on this itself).  I suspect that neither sign nor exp will work well for sparse arrays, so maybe we just drop support for them and use the numpy versions everywhere as well.

pentschev: I'm not sure I understand, are you suggesting we drop all these dispatches that just call `np.some_function()` and call `np.some_function()` directly?

mrocklin: Yes, I think that in many (but not all) cases that will work just fine.  We don't need the dispatch between numpy and dask array.  The `object` dispatch is also often unnecessary.

```python
In [1]: import numpy as np

In [2]: import dask.array as da

In [3]: import scipy.sparse

In [4]: x = scipy.sparse.csr_matrix([[1, 2], [0, 1]])

In [5]: d = da.from_array(x, chunks='auto')

In [6]: np.sign(d)
Out[6]: dask.array<sign, shape=(2, 2), dtype=int64, chunksize=(2, 2)>

In [7]: np.log1p(d)
Out[7]: dask.array<log1p, shape=(2, 2), dtype=float64, chunksize=(2, 2)>

In [8]: abs(d)
Out[8]: dask.array<abs, shape=(2, 2), dtype=int64, chunksize=(2, 2)>

In [9]: np.dot(d, d)  # this seems not to work, it seems to return a computed array
Out[9]:
<2x2 sparse matrix of type '<class 'numpy.int64'>'
	with 3 stored elements in Compressed Sparse Row format>
```

mrocklin:
```diff
-@dispatch(object)
-def log1p(A):
-    return A.log1p()


-@dispatch(np.ndarray)
-def log1p(A):
-    return np.log1p(A)


-@dispatch(da.Array)
-def log1p(A):
-    return da.log1p(A)

+from numpy import log1p
```

Or, probably better, just use `np.log1p` in the code rather than `utils.log1p`

pentschev: I see, I'll check the ones we can get rid of against the current test set we have in-place.

pentschev: I just attempted removing all `dot()` dispatchers and running tests, everything passed. I'm only not sure whether we should keep some specializations for Dask arrays, like this one:

```python
@dispatch(da.Array, np.ndarray)
def dot(A, B):
    B = da.from_array(B, chunks=B.shape)
    return da.dot(A, B)
```

Since this is using Dask chunks, performance may be negatively impacted on cases that do take advantage of this. Any thoughts?

mrocklin: We want to keep the dot dispatch

```python
In [9]: np.dot(d, d)  # this seems not to work, it seems to return a computed array
Out[9]:
<2x2 sparse matrix of type '<class 'numpy.int64'>'
	with 3 stored elements in Compressed Sparse Row format>
```

This failed because it returned a computed value rather than a lazy dask array value.  If this succeeded we would have seen an output like 

    <dask.array<dot, shape=(2, 2), dtype=int64, chunksize=(2, 2)>

pentschev: Alright, I'll leave the `dot()` dispatchers as they are now.

TomAugspurger:
Hmm looks like I dropped the ball on deprecating the estimators here... I
don't really care where we do that. If people still think it makes sense
for the estimators to live in dask-ml, and the optimizers here, then I'll
update that PR.

pentschev:
FWIW, I have no strong opinions there.

Sorry if this question has already been answered before or sounds a bit stupid, but wouldn't it then make sense to have Dask-GLM completely deprecated and move everything into Dask-ML? GLM being a statistical model, it doesn't fall too far off of ML, but of course, maybe syntactically it could be considered something very unrelated.

jrbourbeau:
My personal preference would be to merge #66 and deprecate `dask_glm.estimators` in favor of `dask_ml.linear_model`. That way all the scikit-learn style estimators live in one place.

TomAugspurger:
I updated #66 to fix the merge conflict.

matthieubulte:
Hi, this PR would be quite useful for me.

Are you only waiting for #66 to be finalized and merged to merge this one?

It seems that #66 is paused for now, is that something where I could help to unblock this PR?

Otherwise, anything else I could do to get this merged?

Thanks!

pentschev:
As far as I recall, this PR should mostly work (if not entirely). It's not necessarily dependant upon #66, even though a little modification would be required once it's merged. However, the changes here are dependant upon several unreleased changes:

1. https://github.com/numpy/numpy/pull/13046 -- will be available in NumPy 1.17
1. https://github.com/cupy/cupy/pull/2171 -- will only be merged once NumPy 1.17 is released, then available on the next CuPy release after that
1. https://github.com/cupy/cupy/pull/2079 / https://github.com/cupy/cupy/pull/2219 -- available in CuPy 7.0.0b1, but requires setting `CUPY_EXPERIMENTAL_SLICE_COPY=1` environment variable

That said, we are likely to wait to have all these changes released before we can merge that. I'm expecting that NumPy is released sometime this month, following that, probably a CuPy release with necessary changes is probably due in August or September.

matthieubulte:
Great, thanks for the clarifications!

Issue #77, opened 2019-03-12
Estimators don't have n_iter_ parameter


stsievert:
Scikit-learn estimators have a `n_iter_` parameter. Dask-GLM doesn't.

For example, the Scikit-Learn logistic regression implementation has the `n_iter_` parameter ([docs][sklearn]). However, the Dask-GLM implementation doesn't ([docs][dask]).

[sklearn]:https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html
[dask]:http://ml.dask.org/modules/generated/dask_ml.linear_model.LogisticRegression.html

PR #79, opened 2019-03-20
Add n_iter_ attribute to estimators

pentschev:
Solves #77.

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -98,6 +98,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     recalcRate = 10
     backtrackMult = firstBacktrackMult
     beta = np.zeros(p)
+    n_iter = 0
 
     for k in range(max_iter):
         # how necessary is this recalculation?
@@ -125,6 +126,8 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
         beta = beta - stepSize * grad  # tiny bit of repeat work here to avoid communication
         Xbeta = Xbeta - stepSize * Xgradient
 
+        n_iter += 1
+
         if stepSize == 0:
             break
 
@@ -136,7 +139,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
         stepSize *= stepGrowth
         backtrackMult = nextBacktrackMult
 
-    return beta
+    return beta, n_iter
 
 
 @normalize
@@ -164,7 +167,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     beta = np.zeros(p)  # always init to zeros?
     Xbeta = dot(X, beta)
 
-    iter_count = 0
+    n_iter = 0
     converged = False
 
     while not converged:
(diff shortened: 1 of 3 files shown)


mrocklin:
cc @jrbourbeau @cicdw

dask_glm/algorithms.py
```diff
@@ -181,17 +186,16 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
...
-
         # should change this criterion
         coef_change = np.absolute(beta_old - beta)
         converged = (
-            (not np.any(coef_change > tol)) or (iter_count > max_iter))
+            (not np.any(coef_change > tol)) or (n_iter >= max_iter))
 
         if not converged:
             Xbeta = dot(X, beta)  # numpy -> dask converstion of beta
+            n_iter += 1
```
jrbourbeau: Should `n_iter += 1` be moved to the line after `beta = (beta_old - step)`? Feel free to correct me if I'm wrong, but I think that in the limiting case that the convergence condition is met after a single iteration, this will return `n_iter = 0` right now

pentschev: Hmm, but would that be consider an iteration if no work has been done (e.g., convergence was achieved without work)? I honestly didn't consider that this could happen.

pentschev:
> Minor comment: could you please add an `n_iter_` attribute to the estimator docstrings

Yes, thanks for pointing that I forgot it.

> Also, as you mentioned, these changes will cause breakage in `dask-ml`. Do you have an interest in updating the estimators over there too? Definitely no obligation though, just wanted to check

Yes, I planned to do that once this was merged, I think there's no point in already opening an PR there before this gets reviewed here.

dask_glm/algorithms.py
jrbourbeau: I think that at least one iteration is always done when this function is called. `n_iter` starts out as 0 with `converged = False`

https://github.com/dask/dask-glm/blob/34766b23bfec90a51e7afeaeddfce6d9f096957c/dask_glm/algorithms.py#L172-L173

then we enter the `while` loop and work is done to get an updated value for `beta`

https://github.com/dask/dask-glm/blob/34766b23bfec90a51e7afeaeddfce6d9f096957c/dask_glm/algorithms.py#L182-L187

We then check to see if things have converged 

https://github.com/dask/dask-glm/blob/34766b23bfec90a51e7afeaeddfce6d9f096957c/dask_glm/algorithms.py#L190-L192

Let's pretend that the first time line 191 is evaluated we've converged and now `converged = True`. Now we won't go into the code block that increments `n_iter`

https://github.com/dask/dask-glm/blob/34766b23bfec90a51e7afeaeddfce6d9f096957c/dask_glm/algorithms.py#L194-L196

then we'll break out of the `while` loop and `n_iter` will be returned as `0` even though we've already done one iteration. `n_iter` being off by one should happen for any number of iterations (e.g. `n_iter` being returned as 5 after 6 iterations), it's just easier for me to think about the single iteration case.

jrbourbeau:
> Yes, I planned to do that once this was merged

Awesome, thank you @pentschev

dask_glm/algorithms.py
pentschev: Ok, I see what you mean and I think you're right, my bad. Fixing.

pentschev:
From my side, this is ready for another review/merge.

pentschev:
Are we good to merge this?

PR #80, opened 2019-09-10
Scatter lbfgs weights to workers. Inject dask client to lbfgs.

jdlesage:
In case of optimization in large space (say a hashing space of 2**25 size).
The results can be big (> 55MB) and must be broadcasted to workers.

Note: If the beta array is not scattered, it will be serialized with dask task. It will make the dask graph to big to be handle by the scheduler.

This commit introduce a new option to inject dask distributed client to lbfgs. If present, we use it to scatter the beta vector. In dask-ml, this client can be injected using the solver_kwargs argument.

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -341,8 +341,8 @@ def lbfgs(X, y, regularizer=None, lamduh=1.0, max_iter=100, tol=1e-4,
     def compute_loss_grad(beta, X, y):
         scatter_beta = scatter_array(
             beta, dask_distributed_client) if dask_distributed_client else beta
-        loss_fn = pointwise_loss(beta, X, y)
-        gradient_fn = pointwise_gradient(beta, X, y)
+        loss_fn = pointwise_loss(scatter_beta, X, y)
+        gradient_fn = pointwise_gradient(scatter_beta, X, y)
         loss, gradient = compute(loss_fn, gradient_fn)
         return loss, gradient.copy()
 
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -11,7 +11,7 @@ import dask.array as da
 from scipy.optimize import fmin_l_bfgs_b
 
 
-from dask_glm.utils import dot, normalize, scatter_array
+from dask_glm.utils import dot, normalize, scatter_array, get_distributed_client
 from dask_glm.families import Logistic
 from dask_glm.regularizers import Regularizer
 
@@ -304,8 +304,7 @@ def local_update(X, y, beta, z, u, rho, f, fprime, solver=fmin_l_bfgs_b):
 
 @normalize
 def lbfgs(X, y, regularizer=None, lamduh=1.0, max_iter=100, tol=1e-4,
-          family=Logistic, verbose=False, dask_distributed_client=None,
-          **kwargs):
+          family=Logistic, verbose=False, **kwargs):
     """L-BFGS solver using scipy.optimize implementation
 
     Parameters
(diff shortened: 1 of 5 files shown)


dask_glm/algorithms.py
```diff
@@ -304,7 +304,8 @@ def local_update(X, y, beta, z, u, rho, f, fprime, solver=fmin_l_bfgs_b):
 
 @normalize
 def lbfgs(X, y, regularizer=None, lamduh=1.0, max_iter=100, tol=1e-4,
-          family=Logistic, verbose=False, **kwargs):
+          family=Logistic, verbose=False, dask_distributed_client=None,
```
TomAugspurger: Typically we would call this `client` if we're taking a Client or `scheduler_address`, if we're expecting a string address.

TomAugspurger:
Also, note that distributed now as Actors https://distributed.dask.org/en/latest/actors.html, which may be a better fit for this type of operation.

I think it's OK to take on distributed as a required dependency for these algorithms. I think the single-machine case is pretty well covered by other libraries.

TomAugspurger:
Thanks for the PR by the way! Sorry for the slow response.

dask_glm/algorithms.py
mrocklin: We might also call `dask.distributed.get_client()` and rely on the global.  This is probably easier on users and unlikely to go wrong.  If we are in a situation where there are many clients (very rare) then the user could still control things with `dask.config.set(scheduler=client)`

mrocklin: We could also use the `dask.distributed.get_client()` global for this.

```python
def compute_step(..., scatter=False):
    if scatter:
        from dask.distributed import get_client
        client = get_client()
        beta = client.scatter(beta)
    ...
```

This isn't as explicit, but I suspect that it will almost always be fine, and if someone has multiple clients (this is very rare) then they can control this with `dask.config.set(scheduler=client)`.

mrocklin:
+1 on experimenting with Actors.  I'm not sure that they are appropriate for LBFGS, but in general it would be fun to explore more asynchronous optimization algorithms with actors.

dask_glm/algorithms.py
jdlesage: I did the change using the dask.distributed method + add distributed in requirements

jdlesage:
I will test actors in another commit in this PR. I didn't know this feature in dask... Thanks for the advises!

dask_glm/utils.py
```diff
@@ -3,6 +3,7 @@
 import inspect
 import sys
 
+import dask.distributed as dd
```
TomAugspurger: `dd` is the usual alias for `dask.dataframe`. Could you change this to an `import distributed` or `from distributed import Client?

```suggestion
from distributed import Client
```

jdlesage: Agree I will do. 
I did the change with actors and ran unit tests on it. It is very simple to use. I would like to do a full run on the cluster here to validate my change before to push the commit.

jdlesage:
Version using actors is very slow... I tested with 5 iterations of lbfgs on my data, it usually takes 15 min. But with actors, I stopped the run after 30 min
I will not push this change in this PR. For your information, that's the commit that introduce actors: https://github.com/jdlesage/dask-glm/commit/d847c9851fd2248240a5a411d420763491548330

I suspect it is very slow when the loop reads array metadata (such as shape). If my understanding is correct, as dtype and shape are attributes, it will download all the array to get these info. dask array seems necessary here to have a quick access to the array metadata.

TomAugspurger:
Thanks for trying out an actor version. I'll take a closer look later.

Merging this when it passes.

TomAugspurger:
Thanks @jdlesage!

PR #82, opened 2019-09-18
Add support of sparse matrix

jdlesage:
This commit enable to give sparse matrix (from this lib https://github.com/pydata/sparse/) to the dask-glm estimators

It is linked to this "epic" ticket: https://github.com/dask/dask-ml/issues/123

The changes are:
* Fit intersect handle sparse matrix
* Add option to inject if estimator can manage sparse matrix

Some remarks about the choices:
* I don't know how to detect dynamically if a dask array is sparse. A solution could be to compute 1 task and check the resulting array... It will trigger unnecessary computations. That's why, I decided to pass an option to tell to estimators that matrix will be sparse. 
* normalize will densify the matrix (and raise a Exception in sparse library code). I decided to disable the option when matrix are sparse.
* sparse prevents to mix sparse and dense matrix in concatenate. So I specialized add_intersecept for sparse matrix.

### DIFF
--- a/dask_glm/tests/test_utils.py
+++ b/dask_glm/tests/test_utils.py
@@ -72,8 +72,9 @@ def test_add_intercept_dask():
 
 
 def test_add_intercept_sparse():
+    from sparse.utils import assert_eq
     X = sparse.COO(np.zeros((4, 4)))
-    result = utils.add_sparse_intercept(X)
+    result = utils.add_intercept(X)
     expected = sparse.COO(np.array([
         [0, 0, 0, 0, 1],
         [0, 0, 0, 0, 1],
@@ -85,7 +86,7 @@ def test_add_intercept_sparse():
 
 def test_add_intercept_sparse_dask():
     X = da.from_array(sparse.COO(np.zeros((4, 4))), chunks=(2, 4))
-    result = utils.add_sparse_intercept(X)
+    result = utils.add_intercept(X)
     expected = da.from_array(sparse.COO(np.array([
         [0, 0, 0, 0, 1],
         [0, 0, 0, 0, 1],
@@ -96,7 +97,6 @@ def test_add_intercept_sparse_dask():
 
 
 def test_sparse():
-    sparse = pytest.importorskip('sparse')
     from sparse.utils import assert_eq
     x = sparse.COO({(0, 0): 1, (1, 2): 2, (2, 1): 3})
     y = x.todense()
--- a/dask_glm/tests/test_utils.py
+++ b/dask_glm/tests/test_utils.py
@@ -72,7 +72,6 @@ def test_add_intercept_dask():
 
 
 def test_add_intercept_sparse():
-    from sparse._utils import assert_eq
     X = sparse.COO(np.zeros((4, 4)))
     result = utils.add_intercept(X)
     expected = sparse.COO(np.array([
@@ -81,7 +80,7 @@ def test_add_intercept_sparse():
         [0, 0, 0, 0, 1],
         [0, 0, 0, 0, 1],
     ], dtype=X.dtype))
-    assert_eq(result, expected)
+    assert (result == expected).all()
 
 
 def test_add_intercept_sparse_dask():
@@ -97,9 +96,15 @@ def test_add_intercept_sparse_dask():
 
 
 def test_sparse():
-    from sparse._utils import assert_eq
     x = sparse.COO({(0, 0): 1, (1, 2): 2, (2, 1): 3})
     y = x.todense()
     assert utils.sum(x) == utils.sum(x.todense())
     for func in [utils.sigmoid, utils.sum, utils.exp]:
-        assert_eq(func(x), func(y))
+        assert (func(x) == func(y)).all()
+
+
+def test_dask_array_is_sparse():
+    assert utils.is_dask_array_sparse(da.from_array(
+        sparse.COO([], [], shape=(10, 10))))
+    assert utils.is_dask_array_sparse(da.from_array(sparse.eye(10)))
+    assert not utils.is_dask_array_sparse(da.from_array(np.eye(10)))
--- a/dask_glm/tests/test_utils.py
+++ b/dask_glm/tests/test_utils.py
@@ -108,3 +108,8 @@ def test_dask_array_is_sparse():
         sparse.COO([], [], shape=(10, 10))))
     assert utils.is_dask_array_sparse(da.from_array(sparse.eye(10)))
     assert not utils.is_dask_array_sparse(da.from_array(np.eye(10)))
+
+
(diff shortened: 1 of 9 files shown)


TomAugspurger:
> I don't know how to detect dynamically if a dask array is sparse.

Recent versions of dask have a `._meta` attribute storing this information:

```python
In [8]: X = da.ones((10, 10), chunks=5).map_blocks(sparse.COO)

In [9]: X._meta
Out[9]: <COO: shape=(0, 0), dtype=float64, nnz=0, fill_value=0.0>
```

We can dispatch on that. That removes the need for the `use_sparse_matrix` keyword, right?

Disallowing normalization for now makes sense.

jdlesage:
Great! I will try this _meta

TomAugspurger:
This all looks nice.

How do things seem in practice? Do the algorithms seem to work well with sparse arrays?

jdlesage:
I am testing dask on a dataset similar to the 1TB criteo dataset (I am from Criteo actually...) I hash the dataset with the hashing trick (described here http://olivier.chapelle.cc/pub/ngdstone.pdf) in a hash space of 2**25 length As far as I see, it scales (~2 min on 180 workers to run 1 iteration of LBFGS). That's nice :)
To be 100% honest, I did have time to test only on a small subset of features... but that's promising. 
If I compare with the other Logistic Regression implementations we use here, scipy lbfgs converges slowly (in term of number of iterations) ... I still have to investigate the differences.

jdlesage:
Unit tests are red on travis. I see the sparse lib version differs with my dev environment. I will have a look to fix it.

jdlesage:
travis still uses an old version of sparse. I will try it on a local travis running on docker before to push the next fix.

TomAugspurger:
Strange. I'm going to push a commit to get some debug info on what's installed..

jdlesage:
I use venv instead of conda. I was able to reproduce the problem with a conda environement. Numpy version differs. (1.15.0 for conda vs 1.17.2 for pip). If I downgrade numpy, in my pip env I got the same error.
I tried to upgrade numpy at conda env creation but it is still at 1.15.0. Maybe instead of upgrading numpy, I will investigate to understand why numpy changes sparse behavior and propose a fix.

TomAugspurger:
> Maybe instead of upgrading numpy, I will investigate to understand why numpy changes sparse behavior and propose a fix.

I'd recommend requiring NumPy>=1.17. Most likely it's relying on NumPy's `__array_function__` interface (new in 1.17) to ensure that NumPy functions called on sparse.COO objects use the sparse implementation.

dask_glm/tests/test_utils.py
```diff
@@ -70,9 +71,33 @@ def test_add_intercept_dask():
     assert_eq(result, expected)
 
 
+def test_add_intercept_sparse():
+    from sparse._utils import assert_eq
```
TomAugspurger: @hameerabbasi is there a public API to assert that two sparse arrays are equal?

hameerabbasi: Is all you're using `COO`? In general, for future compat as well, I'd prefer you used `(a == b).all()`.

dask_glm/utils.py
```diff
@@ -137,18 +138,32 @@ def sum(A):
...
+
+
 @dispatch(np.ndarray)
 def add_intercept(X):
     return np.concatenate([X, np.ones((X.shape[0], 1))], axis=1)
 
 
+@dispatch((sparse.COO, sparse.DOK))
+def add_intercept(X):
+    return sparse.concatenate([X, sparse.COO(np.ones((X.shape[0], 1)))], axis=1)
```
TomAugspurger: Is there any advantage to making the `ones` we create the same sparse format as `X`? Right now, we may concat a COO to a DOK, which IIUC results in a COO.

dask_glm/tests/test_utils.py
TomAugspurger: Looks like we're supporting DOK too. But I'm not sure whether we care about the exact output format (maybe @jdlesage has thoughts).

hameerabbasi: I can add some support to `DOK` for NumPy operators and then that'd be all that's needed.

hameerabbasi: `DOK` is very primitive, I'd recommend not supporting it for now.

dask_glm/utils.py
hameerabbasi: `concatenate(DOK, DOK)` is also `COO` so I think this is fine.

hameerabbasi: `concatenate([all_DOK])` is also `COO`, so I'd recommend using `COO`.

dask_glm/utils.py
```diff
@@ -137,18 +138,32 @@ def sum(A):
     return A.sum()
 
 
+def is_dask_array_sparse(X):
+    """
+    Check using _meta if a dask array contains sparse arrays
+    """
+    return isinstance(X._meta, sparse.COO) or isinstance(X._meta, sparse.DOK)
```
jdlesage: Following the discussion on deprecation of `DOK`. Should I test only if the matrix is `COO` ? 
I would like to support only `COO` instead of sparse matrix that is too vague. Wdyt @TomAugspurger  ?

hameerabbasi: Try `sparse.SparseArray`. That's best for forward compat.

dask_glm/utils.py
```diff
@@ -137,18 +138,32 @@ def sum(A):
...
+    """
+    return isinstance(X._meta, sparse.SparseArray)
+
+
 @dispatch(np.ndarray)
 def add_intercept(X):
     return np.concatenate([X, np.ones((X.shape[0], 1))], axis=1)
 
 
+@dispatch((sparse.COO, sparse.DOK))
```
hameerabbasi: This should also ideally be `sparse.SparseArray`.

dask_glm/tests/test_utils.py
```diff
@@ -70,11 +71,40 @@ def test_add_intercept_dask():
...
     for func in [utils.sigmoid, utils.sum, utils.exp]:
-        assert_eq(func(x), func(y))
+        assert (func(x) == func(y)).all()
+
+
+def test_dask_array_is_sparse():
+    assert utils.is_dask_array_sparse(da.from_array(
+        sparse.COO([], [], shape=(10, 10))))
+    assert utils.is_dask_array_sparse(da.from_array(sparse.eye(10)))
+    assert not utils.is_dask_array_sparse(da.from_array(np.eye(10)))
```
jdlesage: FYI: This test does not work with `DOK`. I created an issue in sparse project: https://github.com/pydata/sparse/issues/292
So currently only `COO` matrices are supported.

hameerabbasi: I think an `xfail` would be better, and reference the issue.

TomAugspurger:
Thanks @jdlesage!

PR #85, opened 2020-09-21
Fix the util function "scatter_array" to support both numpy and cupy

daxiongshu:
The current implementation of `dask_glm.utils.scatter_array` always returns a dask array with chunk type `numpy.ndarray` even when the input array is a cupy array. By providing the `meta` argument for `from_delayed`, the returned chunk type matches the input array. No `import cupy` for this PR.


jakirkham:
@TomAugspurger, would you be able to take a look at this? 🙂

TomAugspurger:
Thanks.

PR #87, opened 2020-09-26
Fix algorithms "newton", "gradient_descen" and "proximal_grad" with cupy inputs

daxiongshu:
This PR introduces minor changes that should have no effect on existing functionalities but they enable `cupy` inputs immediately for `dask-glm.algorithms` "newton", "gradient_descen" and "proximal_grad".

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -97,7 +97,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros(p)
+    beta = np.zeros_like(X._meta, shape=(p))
 
     for k in range(max_iter):
         # how necessary is this recalculation?
@@ -161,9 +161,9 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     """
     gradient, hessian = family.gradient, family.hessian
     n, p = X.shape
-    beta = np.zeros(p)  # always init to zeros?
+    beta = np.zeros_like(X._meta, shape=(p))
     Xbeta = dot(X, beta)
-
+    print(Xbeta)
     iter_count = 0
     converged = False
 
@@ -387,7 +387,7 @@ def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros(p)
+    beta = np.zeros_like(X._meta, shape=(p))
     regularizer = Regularizer.get(regularizer)
 
     for k in range(max_iter):
@@ -406,8 +406,8 @@ def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
         # Compute the step size
         lf = func
         for ii in range(100):
-            beta = regularizer.proximal_operator(obeta - stepSize * gradient, stepSize * lamduh)
-            step = obeta - beta
+            beta = regularizer.proximal_operator(- stepSize * gradient + obeta, stepSize * lamduh)
+            step = - beta + obeta
             Xbeta = X.dot(beta)
 
             Xbeta, beta = persist(Xbeta, beta)
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -163,6 +163,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     n, p = X.shape
     beta = np.zeros_like(X._meta, shape=(p))
     Xbeta = dot(X, beta)
+
     iter_count = 0
     converged = False
 
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -97,7 +97,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros_like(X._meta, shape=(p))
+    beta = np.zeros_like(X._meta, shape=p)
 
     for k in range(max_iter):
         # how necessary is this recalculation?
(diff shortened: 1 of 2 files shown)


daxiongshu:
`fit_intercept=True` doesn't work with cupy for now. will work on it.

stsievert:
What do you mean with "it doesn't work"? It throws errors or the expected performance is lower than expected?

daxiongshu:
It throws an error in this function, Line 148 specifically. When `X` is a `dask array with cupy chunks`, `o` is a `dask array with numpy chunks` so `da.concatenate([X, o], axis=1)` throws an error.
https://github.com/dask/dask-glm/blob/69a947e6220c4acaf38aac7181b7d74d947c097a/dask_glm/utils.py#L137-L149

Error message:
```
~/rapids/daskml_cupy/dask-glm/dask_glm/utils.py in add_intercept(X)
    146     # TODO: Needed this `.rechunk` for the solver to work
    147     # Is this OK / correct?
--> 148     X_i = da.concatenate([X, o], axis=1).rechunk((j, (k[0] + 1,)))
    149     return X_i
    150 

~/anaconda3/envs/rapids0.15/lib/python3.7/site-packages/dask/array/core.py in concatenate(seq, axis, allow_unknown_chunksizes)
   3497         type(max(seq_metas, key=lambda x: getattr(x, "__array_priority__", 0)))
   3498     )
-> 3499     meta = _concatenate(seq_metas, axis=axis)
   3500 
   3501     # Promote types to match meta

~/anaconda3/envs/rapids0.15/lib/python3.7/site-packages/cupy/manipulation/join.py in concatenate(tup, axis)
     52         tup = [m.ravel() for m in tup]
     53         axis = 0
---> 54     return core.concatenate_method(tup, axis)
     55 
     56 

cupy/core/_routines_manipulation.pyx in cupy.core._routines_manipulation.concatenate_method()

cupy/core/_routines_manipulation.pyx in cupy.core._routines_manipulation.concatenate_method()

TypeError: Only cupy arrays can be concatenated

```

daxiongshu:
I'll wait this PR to get in first: https://github.com/dask/dask/pull/6680

GenevieveBuckley:
> I'll wait for this PR to get in first: [dask/dask#6680](https://github.com/dask/dask/pull/6680)

Why wait? I don't think my draft PR impacts this one, plus my thing causes a bunch of other stuff to break so there's likely to be a long discussion about what we should do instead. If this is working, I'd encourage you to get it in.

daxiongshu:
Thank you for the comment. This PR depends on `da.ones_like` working correctly with `cupy` input to replace the line `o = da.ones((X.shape[0], 1), chunks=(j, 1))` . If `da.ones_like` with `cupy` is a long shot, I'll try to work around it. But I still hope your PR could get in. Thank you!

GenevieveBuckley:
> This PR depends on `da.ones_like` working correctly with `cupy` input to replace the line `o = da.ones((X.shape[0], 1), chunks=(j, 1))` .

Where is this line, sorry? (I didn't find it with a quick search all across the dask repository)

daxiongshu:
It is line 143 of `utils.py`:
https://github.com/dask/dask-glm/blob/69a947e6220c4acaf38aac7181b7d74d947c097a/dask_glm/utils.py#L137-L149

GenevieveBuckley:
Ah, thank you!

dask_glm/algorithms.py
```diff
@@ -97,7 +97,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros(p)
+    beta = np.zeros_like(X._meta, shape=(p))
```
mrocklin, line 100:
```suggestion
    beta = np.zeros_like(X._meta, shape=p)
```

I think that we probably don't need the extra parens around `p` here and below

dask_glm/algorithms.py
```diff
@@ -406,8 +406,8 @@ def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
         # Compute the step size
         lf = func
         for ii in range(100):
-            beta = regularizer.proximal_operator(obeta - stepSize * gradient, stepSize * lamduh)
-            step = obeta - beta
+            beta = regularizer.proximal_operator(- stepSize * gradient + obeta, stepSize * lamduh)
+            step = - beta + obeta
```
mrocklin, line 410: I'm curious, why this change?

daxiongshu: When `obeta` is a `cupy array`, `cupy's` operators `+/-` can't handle a `dask array` as the second operands in those cases.

pentschev: Maybe you could do `obeta = da.from_array(obeta)` before the computation?

mrocklin: It doesn't matter either way.  Either way is fine.  Can we report this upstream to cupy?  Presumably they need to return NotImplemented somwhere?

mrocklin:
I've removed the WIP label.  This PR seems good to me.  Any objections to merging?

dask_glm/algorithms.py
pentschev: > It doesn't matter either way. Either way is fine.

I know, just providing a more Dask-like solution.

> Can we report this upstream to cupy? Presumably they need to return NotImplemented somwhere?

I don't know exactly what error we get, but would reporting it change anything? Would returning `NotImplemented` really make things much clearer to bother?

pentschev:
> Any objections to merging?

None from me.

pentschev (approved):

pentschev:
It's great to see it took "just" some 17 months to get everything in place for this, since https://github.com/dask/dask-glm/pull/75 . :)

mrocklin:
Thanks @daxiongshu !  This is in.

mrocklin:
> It's great to see it took "just" some 17 months to get everything in place for this, since #75 . :)

Sometimes you plant a seed and it just takes a while to grow :)

dask_glm/algorithms.py
mrocklin: If a Python object doesn't know what to do with another operand in a binary operator, it can return NotIimplemented (or maybe it's raise?) and then Python will try the other order (calling `other.__rsub__(self)`)

pentschev: Thanks @mrocklin , I didn't know of that. I filed https://github.com/cupy/cupy/issues/4118 to start the discussion on the CuPy side, I'm not sure but maybe there's a technical reason not to do that, so let's wait to hear from maintainers.

PR #89, opened 2020-10-10
[Review] Allow lbfgs and admm with dask cupy inputs

daxiongshu:
This PR allows `lbfgs` solver to accept `cupy` inputs. The main changes are:

-  converting `numpy` array, `beta`, to `cupy` array to accelerate `pointwise_loss` and `pointwise_gradient`

-  converting `cupy` arrays, `loss` and `gradients`, to `numpy` arrays to utilize `scipy.optimize.fmin_l_bfgs_b`

Similar changes are made to add `cupy` support to `admm`.

`dask cupy tests` are added for all 5 solvers.

PR #90, opened 2020-10-29
[WIP] Allow numpy array (not dask array) as inputs

daxiongshu:
Currently `dask_glm.estimators` only accepts `dask.array` as inputs due to this line and other places where `._meta` is accessed without checking the data type.

https://github.com/dask/dask-glm/blob/7b2f85fe043eb29212755e67e33e3df553ed0e58/dask_glm/estimators.py#L67

Is `dask_glm.algorithms` used outside `dask_glm.estimators`?

<details>
  <summary>Click to see the example code and error</summary>

Code:
```
from dask_glm.estimators import LogisticRegression
import numpy
x = numpy.random.rand(10,4)
y = numpy.random.rand(10)

lr = LogisticRegression()
lr.fit(x,y)
```
Error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-e644bf405118> in <module>
----> 1 lr.fit(x,y)

~/rapids/daskml_cupy/dask-glm/dask_glm/estimators.py in fit(self, X, y)
     65         X_ = self._maybe_add_intercept(X)
     66         fit_kwargs = dict(self._fit_kwargs)
---> 67         if is_dask_array_sparse(X):
     68             fit_kwargs['normalize'] = False
     69 

~/rapids/daskml_cupy/dask-glm/dask_glm/utils.py in is_dask_array_sparse(X)
    122     Check using _meta if a dask array contains sparse arrays
    123     """
--> 124     return isinstance(X._meta, sparse.SparseArray)
    125 
    126 

AttributeError: 'numpy.ndarray' object has no attribute '_meta'
```
  

</details>

This PR allows numpy arrays (not dask numpy array) as input directly.

### DIFF
--- a/dask_glm/algorithms.py
+++ b/dask_glm/algorithms.py
@@ -11,7 +11,7 @@ import dask.array as da
 from scipy.optimize import fmin_l_bfgs_b
 
 
-from dask_glm.utils import dot, normalize, scatter_array, get_distributed_client
+from dask_glm.utils import dot, normalize, scatter_array, get_distributed_client, safe_zeros_like
 from dask_glm.families import Logistic
 from dask_glm.regularizers import Regularizer
 
@@ -97,7 +97,7 @@ def gradient_descent(X, y, max_iter=100, tol=1e-14, family=Logistic, **kwargs):
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros_like(X._meta, shape=p)
+    beta = safe_zeros_like(X, shape=p)
 
     for k in range(max_iter):
         # how necessary is this recalculation?
@@ -161,7 +161,7 @@ def newton(X, y, max_iter=50, tol=1e-8, family=Logistic, **kwargs):
     """
     gradient, hessian = family.gradient, family.hessian
     n, p = X.shape
-    beta = np.zeros_like(X._meta, shape=p)
+    beta = safe_zeros_like(X, shape=p)
     Xbeta = dot(X, beta)
 
     iter_count = 0
@@ -387,7 +387,7 @@ def proximal_grad(X, y, regularizer='l1', lamduh=0.1, family=Logistic,
     stepSize = 1.0
     recalcRate = 10
     backtrackMult = firstBacktrackMult
-    beta = np.zeros_like(X._meta, shape=p)
+    beta = safe_zeros_like(X, shape=p)
     regularizer = Regularizer.get(regularizer)
 
     for k in range(max_iter):
--- a/dask_glm/tests/test_estimators.py
+++ b/dask_glm/tests/test_estimators.py
@@ -44,13 +44,13 @@ def test_pr_init(solver):
 
 
 @pytest.mark.parametrize('fit_intercept', [True, False])
-@pytest.mark.parametrize('is_sparse', [True, False])
-@pytest.mark.parametrize('is_numpy', [True, False])
+@pytest.mark.parametrize('is_sparse,is_numpy', [
+                         (True, False),
+                         (False, False),
+                         (False, True)])
 def test_fit(fit_intercept, is_sparse, is_numpy):
     X, y = make_classification(n_samples=100, n_features=5, chunksize=10, is_sparse=is_sparse)
     if is_numpy:
-        if is_sparse:
-            return
         X, y = dask.compute(X, y)
     lr = LogisticRegression(fit_intercept=fit_intercept)
     lr.fit(X, y)
--- a/dask_glm/tests/test_estimators.py
+++ b/dask_glm/tests/test_estimators.py
@@ -45,8 +45,13 @@ def test_pr_init(solver):
 
 @pytest.mark.parametrize('fit_intercept', [True, False])
 @pytest.mark.parametrize('is_sparse', [True, False])
-def test_fit(fit_intercept, is_sparse):
+@pytest.mark.parametrize('is_numpy', [True, False])
+def test_fit(fit_intercept, is_sparse, is_numpy):
     X, y = make_classification(n_samples=100, n_features=5, chunksize=10, is_sparse=is_sparse)
+    if is_numpy:
+        if is_sparse:
+            return
+        X, y = dask.compute(X, y)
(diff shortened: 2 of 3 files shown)


daxiongshu:
@mrocklin @pentschev I just added one test for now. If it is ok, could you please suggest which other tests I should add `numpy input`? Thank you!

daxiongshu:
I think I'm going to finish this first and then move on to #89

dask_glm/algorithms.py
```diff
@@ -11,7 +11,7 @@
 from scipy.optimize import fmin_l_bfgs_b
 
 
-from dask_glm.utils import dot, normalize, scatter_array, get_distributed_client
+from dask_glm.utils import dot, normalize, scatter_array, get_distributed_client, safe_zeros_like
```
pentschev, line 14: Where is `safe_zeros_like` coming from? I suppose you wanted to `from dask.array.utils import zeros_like_safe` instead, from https://github.com/dask/dask/blob/48a4d4a5c5769f6b78881adeb1b3973a950e5f43/dask/array/utils.py#L350

dask_glm/utils.py
```diff
@@ -205,3 +210,9 @@ def get_distributed_client():
         return get_client()
     except ValueError:
         return None
+
+
+def safe_zeros_like(X, shape):
+    if isinstance(X, da.Array):
+        return np.zeros_like(X._meta, shape=shape)
+    return np.zeros_like(X, shape=shape)
```
pentschev, line 218:
```suggestion
    return zeros_like_safe(meta_from_array(X))
```

pentschev: You'll also need to `from dask.array.utils import meta_from_array` at the top.

dask_glm/utils.py
```diff
@@ -149,6 +149,11 @@ def add_intercept(X):
     return X_i
 
 
+@dispatch(object)
+def add_intercept(X):
+    return np.concatenate([X, np.ones_like(X, shape=(X.shape[0], 1))], axis=1)
```
pentschev, line 154:
```suggestion
    return np.concatenate([X, ones_like_safe(X, shape=(X.shape[0], 1))], axis=1)
```

pentschev: Also needs `from dask.array.utils import ones_like_safe`.

dask_glm/tests/test_estimators.py
```diff
@@ -44,9 +44,14 @@ def test_pr_init(solver):
...
+                         (True, False),
+                         (False, False),
+                         (False, True)])
+def test_fit(fit_intercept, is_sparse, is_numpy):
     X, y = make_classification(n_samples=100, n_features=5, chunksize=10, is_sparse=is_sparse)
+    if is_numpy:
+        X, y = dask.compute(X, y)
     lr = LogisticRegression(fit_intercept=fit_intercept)
     lr.fit(X, y)
     lr.predict(X)
```
pentschev, line 57: I don't think I understand this test. When is `is_numpy` the case in a real-world example, IOW, will you ever have `X` and `y` be pure NumPy arrays that's worth testing with `LogisticRegression`? I assumed you'd only have Dask arrays (backed by Sparse or not).

pentschev (changes requested):
@daxiongshu I added a few requests to make the code easier and more Dask-like, also a few questions on things that aren't clear to me. Please take a look when you have a moment.

dask_glm/utils.py
daxiongshu: Sorry for the late reply, I think I might misunderstand our other conversion. This PR intends to enable `dask-glm` to deal with `pure numpy arrays`. Please let me know if not so and `dask-glm` should only accept `dask arrays`.

https://github.com/dask/dask-glm/blob/7b2f85fe043eb29212755e67e33e3df553ed0e58/dask_glm/algorithms.py#L100-L101

Let's say the input `X` is a pure `numpy` or `cupy` array, not a dask array. `beta = np.zeros_like(X._meta)` will be an error. The `safe_zeros_like` (bad naming) I implemented will check if `X` is a `pure numpy/cupy array` or a `dask array` and **return a `pure numpy/cupy array`**.  In contrast, `da.utils.zeros_like_safe` **returns a `dask array`.**  In this case the beta should be a `pure numpy/cupy array`.

Let me know if this clears things up. Thank you!

dask_glm/tests/test_estimators.py
daxiongshu: That's exactly what I tried to do, where both `X` and `y` are pure `numpy/cupy arrays`. Is that a feature we want? The current `dask-glm` only accepts `dask arrays`.

dask_glm/utils.py
pentschev: > The `safe_zeros_like` (bad naming) I implemented will check if `X` is a `pure numpy/cupy array` or a `dask array` and **return a `pure numpy/cupy array`**.

That's exactly what `meta_from_array` does. It will return an array of the type `_meta` has (i.e., _chunk type_), so if the input is a NumPy array or a Dask array backed by NumPy, the result is an empty `numpy.ndarray`, and if the input is a CuPy array or a Dask array backed by CuPy, the result is an empty `cupy.ndarray`.

> In contrast, `da.utils.zeros_like_safe` **returns a `dask array`.**

That isn't necessarily true, it will only return a Dask array if the reference array is a Dask array. Because we're getting the _underlying chunk type_ with `meta_from_array`, the resulting array will be either a NumPy or CuPy array.

dask_glm/tests/test_estimators.py
pentschev: I don't think that's a feature we need to support explicitly, I believe anybody using `dask-glm` would want to use Dask arrays rather than pure NumPy/CuPy ones.

dask_glm/utils.py
daxiongshu: Aha, that works! I will make the changes.

dask_glm/tests/test_estimators.py
daxiongshu: Thank you! I'll prioritize #89 then.

Issue #99, opened 2022-12-06
Archive this repo?


jakirkham:
It appears this is currently unmaintained. If that is the case, should we go ahead and archive this repo?

jakirkham:
cc @cicdw @TomAugspurger

TomAugspurger:
+1On Dec 6, 2022, at 4:49 PM, jakirkham ***@***.***> wrote:﻿
cc @cicdw @TomAugspurger

fjetter:
FYI There is some interest in reviving this repo and we will invest some time to get this up and running again

TomAugspurger:
The source lives at https://github.com/dask/dask-ml/blob/main/dask_ml/linear_model/glm.py and related files now. Or is there some other reason to revive it?

PR #100, opened 2023-09-12
CI: Get CI green after unarchiving repository

milesgranger:


milesgranger:
@fjetter It's green; ready for review. At least one quick follow up would be to address some xfails added here spawning from updating the `sparse` library, then improving some of the CI maybe with pre-commit?

.github/workflows/ci.yml
```diff
@@ -9,38 +9,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
```
fjetter, line 12: `dask/dask` and `dask/distributed` is only supporting 3.9+, I suggest to drop 3.8 right away
```suggestion
        python-version: ["3.9", "3.10", "3.11"]
```

fjetter (approved):
I'm OK to merge once the python version is dropped

fjetter:
> At least one quick follow up would be to address some xfails added here spawning from updating the sparse library, then improving some of the CI maybe with pre-commit?

Yes, I think we should do both

TomAugspurger:
@milesgranger @fjetter just making sure you saw https://github.com/dask/dask-glm/issues/99#issuecomment-1716206031.

>  The source lives at https://github.com/dask/dask-ml/blob/main/dask_ml/linear_model/glm.py and related files now. Or is there some other reason to revive it?

milesgranger:
@TomAugspurger  I could be missing something here, but seems `glm.py` [still imports from dask_glm](https://github.com/dask/dask-ml/blob/b5640cbb913954a227585cae413d89d6b48f4c0f/dask_ml/linear_model/glm.py#L5), and that maybe dask_glm doesn't need to have [estimators.py](https://github.com/dask/dask-glm/blob/main/dask_glm/estimators.py), but would still be relevant to revive for continued maintenance?

TomAugspurger:
Huh, I guess I'm mistaken. I thought that all of the source had been moved over, but apparently it was just the scikit-learn API compatible stuff.
________________________________

PR #106, opened 2023-11-06
Add wheels.yml workflow for release to PyPI

milesgranger:


.github/workflows/wheels.yml
```diff
@@ -0,0 +1,65 @@
+name: Build and maybe upload to PyPI
+
+on:
+  push:
+  pull_request:
```
fjetter, line 5: Does this actually have to run on push/pull_request? I would expect `release` to be sufficient. `push` maybe but only for `main`?

milesgranger: Ya, I did that because the actual uploading does a check for tags being present. 

`if: "startsWith(github.ref, 'refs/tags/')"`

Thought maybe it'd be useful to ensure the project builds successfully, but maybe it's overkill.

fjetter (approved):

